PR #1, opened 2020-07-02
Adds phone number validation. Country of field not stored.

plocket:


purplesky2016:
Looking great! 

Suggestion: since not everyone knows how to call a module crossing interviews, maybe move the yml file to the other project VirtualCourtSampleInterviews, which would allow the full calling code of your module from outside of this project, like this:
---
modules:  
  - docassemble.VirtualCourtToolbox.PhoneNumberDataType

(Other than maintenance concerns, this is why I separated sample interviews from the toolbox project.)

plocket:
Thanks! That isn't a sample interview, actually, though - that is the file the coder should `include:` in order to get that feature. All the code that is there is necessary to make the phone datatype work. With this file, the coder can just pull in that YAML file and it'll do all the work for them. I have an actual file with the example use, I'm just not sure when I should make that PR.

plocket:
Feel free to merge once you're good with what's here, @purplesky2016 !

purplesky2016:
Okay, I'll merge it now.

PR #3, opened 2020-07-09
Remove the behavior of altering user entered dates when invalid

plocket:
Spacing looks strange because the code uses tabs instead of spaces. I can fix that on another go around.

Not sure why git thinks that everything's been replaced.


purplesky2016:
Great! You can give a demo on today's call.

PR #4, opened 2020-07-10
Implement client-side validation for date with CustomDataType

plocket:
Not sure if you wanted this or not. We can eventually move the other javascript into here as well and reduce the number of files.

I wasn't sure whether to delete the function in `.misc`. Let me know your thoughts.

I'd also appreciate your thoughts on removing restrictions on the inputs or, alternatively, figuring out how to replace their error messages with ones for the whole field instead. That last one may be tricky.


plocket:
Ah, conflicts. I'll remove this PR and replace it with one that doesn't conflict.

PR #5, opened 2020-07-10
Validate in `CustomDataType` instead of in YAML. Delete unused file.

plocket:
We can move the js into here as well so that we don't have to include as many files. I just didn't want to do it at the same time.

Also, keeping the date validation function in `.misc` might be a useful toolbox tool, so I left it where it was. Would appreciate thoughts on that.


purplesky2016:
Nice work, Love it!

PR #9, opened 2020-10-01
Change name of repository, inside and out

plocket:


nonprofittechy (approved):
This looks good to me so long as it runs. I don't know best way to test--feel free to merge if you're confident about it.

PR #10, opened 2020-10-02
Added custom datatype, changed the datatype name and moved the exampl…

purplesky2016:
…e yml file to here.


nonprofittechy:
1. This comes in as a text date. How does it work when the user's browser is set to use different date formats?
2. Can we convert it to a DADateTime?

PR #17, opened 2021-05-15
Close #16, share button with accessible text

### LINKED ISSUE #16: Add a function to create a copy to clipboard input
Here is a hardcoded one that is not quite right yet:

```yaml
fields:
  - note: |
      Copy and share this link
      <input disabled="true" type="text" value="${ interview_url(i=user_info().filename, style="short", new_session=1) }" id="myInput">
      <div class="copy-tooltip">
      <button onclick="myFunction()" onmouseout="outFunc()" type="button">
        <span class="tooltiptext" id="myTooltip">Copy to clipboard</span>
        <i class="far fa-copy"></i>
        </button>
      </div>      
script: |
  <script>
    function myFunction() {
      var copyText = document.getElementById("myInput");
      copyText.select();
      copyText.setSelectionRange(0, 99999);
      document.execCommand("copy");

      var tooltip = document.getElementById("myTooltip");
      tooltip.innerHTML = "Copied: " + copyText.value;
    }

    function outFunc() {
      var tooltip = document.getElementById("myTooltip");
      tooltip.innerHTML = "Copy to clipboard";
    }
  </script>
css: |
  <style>
  .copy-tooltip {
    position: relative;
    display: inline-block;
  }

  .copy-tooltip .tooltiptext {
    visibility: hidden;
    width: 140px;
    background-color: #555;
    color: #fff;
    text-align: center;
    border-radius: 6px;
    padding: 5px;
    position: absolute;
    z-index: 1;
    bottom: 150%;
    left: 50%;
    margin-left: -75px;
    opacity: 0;
    transition: opacity 0.3s;
  }

  .copy-tooltip .tooltiptext::after {
    content: "";
    position: absolute;
    top: 100%;
    left: 50%;
    margin-left: -5px;
    border-width: 5px;
    border-style: solid;
    border-color: #555 transparent transparent transparent;
  }

  .copy-tooltip:hover .tooltiptext {
    visibility: visible;
    opacity: 1;
  }
  </style>
```

@plocket is this up your alley?

plocket:
Close #16 from what I understand of the description. Can't see the full function signature in the Playground sidebar unfortunately.


nonprofittechy:
This looks good! Just a suggestion from a quick review of the text:

The name "share button" is not very descriptive of what the function can do. The first use case is on a "sharing" screen, but the input itself does not share anything, and it could be used in other contexts that do not relate to sharing (maybe we want to walk a developer through configuring an API key that we generate, for example). What about "clipboard input" or  "copy button" or some variation that emphasizes the copying and clipboard aspect? 

I recommend ending the function name with `_html` for consistency with other functions that produce HTML output.

docassemble/ALToolbox/share_button.py
```diff
@@ -0,0 +1,30 @@
...
+  text_before='',
+  label='Copy',
+  tooltip_inert_text='Copy to clipboard',
+  tooltip_copied_text='Copied: '  # Will have `text_to_copy` added to it
+):
+  '''Return the html for a button that will let a user share the given text'''
+  button_str = '<div class="al_share">\n'
+  if ( text_before != '' ):
+    button_str += f'<span class="al_share_description">{word( text_before )}</span>\n'
+  button_str += f'<input readonly class="al_share_value" type="text" value="{word( text_to_copy )}">\n'
```
nonprofittechy, line 15: There is no need to run the `text_to_copy` through `word()`

docassemble/ALToolbox/share_button.py
```diff
@@ -0,0 +1,30 @@
+from docassemble.base.functions import word, log
+
+# See GitHub issue https://github.com/SuffolkLITLab/docassemble-ALToolbox/issues/16
+def share_button (
+  text_to_copy,
+  text_before='',
+  label='Copy',
+  tooltip_inert_text='Copy to clipboard',
+  tooltip_copied_text='Copied: '  # Will have `text_to_copy` added to it
+):
```
nonprofittechy, line 10: Suggest adding type annotations as outlined here:

```suggestion
  text_to_copy:str,
  text_before:str='',
  tooltip_inert_text:str='Copy to clipboard',
  tooltip_copied_text:str='Copied: '  # Will have `text_to_copy` added to it
  )->str:
```

docassemble/ALToolbox/share_button.py
```diff
@@ -0,0 +1,30 @@
...
+  button_str += f'<input readonly class="al_share_value" type="text" value="{word( text_to_copy )}">\n'
+  button_str += f'<button class="btn btn-primary al_share_button" type="button">\n'
+  button_str += f'<span class="al_tooltip al_tooltip_inert">{word( tooltip_inert_text )}</span>\n'
+  
+  if ( tooltip_copied_text == 'Copied: ' ):
+    # Our default text will show 'Copied: text that will be copied'
+    button_str += f'<span class="al_tooltip al_tooltip_active">Copied: {word( tooltip_copied_text ) + text_to_copy }</span>\n'
+  else:
+    # Otherwise, use the custom text handed in
+    button_str += f'<span class="al_tooltip al_tooltip_active">Copied: {word( tooltip_copied_text )}</span>\n'
```
nonprofittechy, line 24: This is confusing to me. In both cases it says "Copied" without any translation. And what is the distinction between the two branches? Shouldn't it just look like 
`button_str += f'<span class="al_tooltip al_tooltip_active">{word( tooltip_copied_text ) + text_to_copy }</span>\n'`
in both described circumstances? An option to not show the copied text seems reasonable, but this method of invoking it doesn't make sense to me.

docassemble/ALToolbox/share_button.py
```diff
@@ -0,0 +1,30 @@
...
+  
+  button_str += f'<i class="far fa-copy"></i>\n'
+  button_str += f'<span>{ label }</span>\n'
```
nonprofittechy, line 27: `label` should be passed through `word()`

nonprofittechy (changes requested):
This looks almost ready, and is already a huge improvement over the original code! Just see the inline suggestions.

docassemble/ALToolbox/share_button.py
plocket: Would the option to not show the copied text be another parameter? If I were a dev using this module and handed in something to show in the tooltip called 'tooltip_copied_text', I would expect the value I give to be the one shown in the tooltip after the text had been copied. That is, I wouldn't expect it to be modified.

We could split it up into two parameters perhaps - one for the text before the copied text and one for choosing to not show the copied text. I haven't been able to think of a clear name for the first param, though. `tooltip_text_to_the_left_of_copied_text`? The other could be something like `show_copied_text_in_tooltip` or something.

BryceStevenWilley: IMO, you don't need to show the copied text in the tooltip at all? It's already on the screen and associated with the button. And when the tooltip is really long, the beginning just cuts off, which is much more confusing

Screenshot from 2021-05-17 09-36-21

For names, I think the existing `tooltip_copied_text` is great, and doesn't need to change, if we do want to include the copied text in the tooltip, a good name would be `append_copied_text_to_tooltip`

BryceStevenWilley (approved):
LGTM, don't really think we need to support showing the copied text in the tooltip, but that's just my personal opinion

docassemble/ALToolbox/share_button.py
nonprofittechy: I'd be fine with removing the text that actually gets added to the clipboard from the "Copied!" tooltip. That would simplify this.

plocket:
I _think_ that resolves the current needs. Let me know if I missed something.

nonprofittechy (approved):

nonprofittechy:
Thanks for all of your work on this!

PR #18, opened 2021-05-19
Just bundled up the progressive disclosure CSS that we're using in a …

nonprofittechy:
…few interviews already

This is a tiny tweak from https://docassemble.org/docs/recipes.html#progressive%20disclosure.

This is used in a few AssemblyLine interviews but it hadn't been put into the Toolbox yet.


nonprofittechy:
Going to merge--this is a new, isolated function that doesn't interfere with any existing code.

PR #21, opened 2021-07-14
Added profiling information, so it's not just in Github Issues

BryceStevenWilley:
Info compiled from https://github.com/SuffolkLITLab/docassemble-AssemblyLine/issues/36#issuecomment-862012793 onward in that thread.

Would be interested if someone tries it out with `snakeviz`, curious how it looks. I only know about callgrind and kcachegrind because of my experience with C++.

### DIFF
--- a/docassemble/ALToolbox/data/questions/Profiling.yml
+++ b/docassemble/ALToolbox/data/questions/Profiling.yml
@@ -1,0 +1,63 @@
+---
+comment: |
+  This file helps you determine which parts of your interview take a long time to run.
+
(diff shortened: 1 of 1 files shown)


plocket (approved):
Looking good. I got the file to work on mac:

```bash
python3 -m pip install snakeviz
python3 -m snakeviz test.prof
```

I don't see this logging the url in the logs, but if that's for the next stage of development, then this looks good.

Possible improvement for the future - let people give the file a custom name. That said, people can always re-write that block in their own code.

Issue #38, opened 2021-12-13
Add phone number validation demo


purplesky2016:
As discussed on today's deep dive call.

- add a demo file along with some slight changes of .js and .py. We'll use the tool in some interview to seek feedback. 
- add an entry to the landing page list for this tool.

PR #39, opened 2021-12-13
Phone validation: Fix issue #38

purplesky2016:
@plocket - I didn't change `phone-number-validation.yml `to conform to the name convention, just so that you can see what's changed in the file itself. Feel free to change the name yourself or let me know if you want me to change, or we just leave it alone.

### DIFF
--- a/docassemble/ALToolbox/PhoneNumberDataType.py
+++ b/docassemble/ALToolbox/PhoneNumberDataType.py
@@ -3,7 +3,7 @@ from docassemble.base.util import CustomDataType, DAValidationError
 
 class PhoneNumber( CustomDataType ):
     name = 'phone'
-    input_class = 'dal-phone'
+    input_class = 'al-intl-phone'
     javascript = """\
 /* # Resources
 *  1. What we're using: https://github.com/jackocnr/intl-tel-input
@@ -28,7 +28,7 @@ let validatePhoneNumber = function( value, element, params ) {
 $.validator.addMethod( 'phone', validatePhoneNumber );
 """
     jq_rule = 'phone'
-    jq_message = 'You need to enter a valid phone number.'
+    jq_message = 'Invalid input. Did you enter a "+" before your number?'
     
     # No server-side validation. Just avoiding user error here.
     # If you want to discuss that decision, flag someone else for that.
--- a/docassemble/ALToolbox/PhoneNumberDataType.py
+++ b/docassemble/ALToolbox/PhoneNumberDataType.py
@@ -2,33 +2,49 @@ import re
 from docassemble.base.util import CustomDataType, DAValidationError
 
 class PhoneNumber( CustomDataType ):
-    name = 'phone'
-    input_class = 'al-intl-phone'
+    name = 'al_intl_phone'
+    input_class = 'al_intl_phone'
     javascript = """\
-/* # Resources
-*  1. What we're using: https://github.com/jackocnr/intl-tel-input
-*  1. Especially see: https://github.com/jackocnr/intl-tel-input#static-methods getInstance
-*  1. https://www.npmjs.com/package/google-libphonenumber
-*  1. https://github.com/google/libphonenumber/blob/master/FALSEHOODS.md
-*  1. source: https://www.sitepoint.com/working-phone-numbers-javascript/
+/** When combined with the phone number javascript file shown in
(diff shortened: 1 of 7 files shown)


docassemble/ALToolbox/data/questions/altoolbox_overview.yml
```diff
@@ -24,9 +24,10 @@ subquestion: |
 
   1. Use **snake_case.ext** to name all your files   
   1. Add a demo to show others how to use your tool      
-      - Always include "_demo" at the end of the demo yml file
+      - Always include "_demo" at the end of the demo yml file name
       - In the metadata section add a **title** starting with **ALToolbox - **.
   1. If your tool uses more than one file, add a **package** yml file for easy reference. See "copy_button.yml" as an example.
+      - Do **not** add any metadata in the package file as it will override the interview's matadata.
```
plocket, line 30: Great notes in here. I especially like this one.

docassemble/ALToolbox/data/questions/altoolbox_overview.yml
```diff
@@ -61,6 +62,7 @@ code: |
     "collapse_template_demo": ['Collapse/open text display on the screen', 'collapse_template.yml'],
     "copy_button_demo": ['Add a copy button with tooltip', 'copy_button.yml'],
     "display_template_demo": ['Display the subject and content of a markdown template with/without scroll bar; optional css class', 'display_template.yml'], 
+    "phone_number_validation_demo": ['Validate international phone number input', 'phone-number-validation.yml'],
```
plocket, line 65: Would this be a good place to put a link about how phone number validation can be inconsistent and lead to some people not being able to put in a phone number?

plocket (changes requested):
Before moving forward with this PR, I'd like to discuss what we want to do about validation of phone numbers. It can be problematic if our validator is wrong and this doesn't let someone move forward. Secondly, we might not need to stop someone from moving forwared just if they left out '+'. All seems to do is make the country flag wrong. Right now it actually doesn't seem to stop someone from moving forward with an invalid US number, and maybe that's an ok way to handle it.

Other than that discussion and some fixes, this PR will move this datatype forward. Thanks for taking this on.

On to some of the fixes that are needed:

Right now, the fields are just plain input fields and the validation isn't triggered. That's because some of the `question` values are mismatched with the javascript selectors. I made various comments and suggestions, but I'll try to implement them myself in a new branch. There may also be some stuff that I missed in here. Then I'll make a PR on your branch so you/we can review them before pulling them in here.

If I take too long, or if you just want to implement the changes yourself, feel free to go ahead and make those, but do let me know so we're not doubling up on our work.

docassemble/ALToolbox/data/questions/altoolbox_overview.yml
plocket: Maybe it's better to put that in the text of the demo. I'm not sure.

plocket:
We might want to add a future goal to use `getValidationError` of the phone number validation library (https://www.npmjs.com/package/intl-tel-input#user-content-public-methods), along with the `validate` customDataType class method (https://docassemble.org/docs/fields.html#custom%20datatype) to give better error messages to the user.

We could also try to manage client-side validation so they don't have to press 'continue' to get the validation message.

docassemble/ALToolbox/PhoneNumberDataType.py
```diff
@@ -28,7 +28,7 @@ class PhoneNumber( CustomDataType ):
 $.validator.addMethod( 'phone', validatePhoneNumber );
 """
     jq_rule = 'phone'
-    jq_message = 'You need to enter a valid phone number.'
+    jq_message = 'Invalid input. Did you enter a "+" before your number?'
```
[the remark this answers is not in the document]
plocket: This message is a bit of a problem if it confuses people from the US. If they add a '+' to the front of a correct number, their country becomes egypt or something like that.

I can see this happening when a person types in an invalid number, gets that message, incorrectly adds the plus. They then get the message again because their number is still invalid, so they fix other parts of it and then hit 'submit' without removing the '+'.

Users do crazy things, but that's who we're building it for. That said, we may need to leave this detail for future people to work on.

plocket:
I do want to change the name of the file, but only once we're done reviewing the changes to the code.

PR #43, opened 2022-02-02
Business Days - Fix #41

### LINKED ISSUE #41: Add a business day function to help with date math
Should use the `holidays` Python module as well as accounting for Saturday and Sunday

purplesky2016:
@nonprofittechy - I added a note in the .py file just in case others encounter it, because it took me a while to understand what happened. You can test it by changing the input in the .yml file to see what I mean. 

I don't think we need to alert the user about it, but do we need to consider "what if someone happens to test it that way"?


docassemble/ALToolbox/business_days.py
```diff
@@ -0,0 +1,87 @@
...
+      12/31/2021: New Year’s Day (Observed) and 01/01/2022: New Year’s Day 
+  are both included in 2022 holidays. If remove_holidays=['new year'], then after this line of code:
+      local_holidays.pop_named(record) 
+  the results will erroneously include 2021's holidays as well. But practically no one would delete New Year anyway, we should be fine and there is no need to alarm the user about this issue.
+"""
+def non_business_days(date, interval=0, country='US', state='MA', province=None, add_holidays=None, remove_holidays=None, first_n_dates=0) -> dict:      
+  # 1. Collect weekends, standard holidays and user defined add/remove holidays
+  # 1.1 Get all saturdays and sundays in the given year  
+  date = as_datetime(date)
+  year = date.year
```
BryceStevenWilley, line 22: How do you see this being used on year boundaries? I see this API mostly being used to get which days courts are available in the next 2-3 months, so it might make sense if the date is in November or December to also generate the dates in the next year as well.

BryceStevenWilley, line 22: A related but different question to year boundaries is if we need to pass in the date at all? It's only using the year of the date, so would it make more sense to just directly pass the year (or years if we want to support generating multiple years of holidays?)

BryceStevenWilley (commented):
Overall this is great, but I did realize a potential API change that we might want to consider regarding if we pass the date or just the year to get holidays for, and if we want to support generating dates for multiple years.

docassemble/ALToolbox/business_days.py
BryceStevenWilley: I'm realizing that the holidays to add also contain the year, so that might make it more difficult to generate holidays into the next year (and in general, increase the maintenance cost for keeping hardcoded holidays up-to-date as the years go on). Maybe accepting dates not as strings but as general year-independent rules (like the `holidays` [package does internally](https://github.com/dr-prodigy/python-holidays/blob/7ced8bf6235b17fbe57b69304c4537426653639e/holidays/countries/united_states.py#L178)) would make sense.

docassemble/ALToolbox/business_days.py
```diff
@@ -0,0 +1,87 @@
...
+    return False
+  else: 
+    return True
+  
+def get_next_business_day(date, interval=0, country='US', state='MA', province=None, add_holidays=None, remove_holidays=None) -> dt:	
+  index = interval    
+  done = False
+  while not done: 
+    new_date = (as_datetime(date) + datetime.timedelta(days=index)).strftime('%m/%d/%Y')
+    done = is_business_day(date=new_date, country=country, state=state, province=province, add_holidays=add_holidays, remove_holidays=remove_holidays)
```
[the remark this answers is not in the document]
purplesky2016: Yes, it would do for the boolean var, but we also need to return a new date, thus we need two separate vars.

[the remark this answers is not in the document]
BryceStevenWilley: Not sure how I missed that, lol. Sorry

docassemble/ALToolbox/business_days.py
purplesky2016: Good comments! 

Waiting for a future version update sounds good to me, as we may also collect some feedback from users by then. I'll leave this conversation open so we can revisit when the time comes.

purplesky2016: Strictly speaking, you are right that only the year matters for this particular function. 

However, there are three functions in this module with similar inputs, for user's sake it might be easier to keep the input items consistent for all these functions, as described in the intro page. What do you think?

BryceStevenWilley: I think either works fine: for the possible extension we discussed earlier, you can auto get holidays for the next year by looking at the date, so you're guaranteed holidays for 60 days out, or you could pass the years, so you always get all holidays for all the years you requested. I'll leave it up to your judgement, which seems to be changing it to year.

purplesky2016: I just discovered that `holidays `module had a major change about 2 weeks ago, the readme page changed, and the function I am using (CountryHoliday) is deprecated! What a perfect timing! https://github.com/dr-prodigy/python-holidays/commit/18f02630520ea7da29c86200ce8143cb6dac0a26. Oh well, better now than after we released our code. I'll make the switch.
image

purplesky2016: On a positive note, I tested some cross-year scenarios using the current code and they worked. I think it's because we loop thru each day and call non_business_days for it individually, which means we are always dealing with a "current day", thus no worries about year boundary. 

I am glad we are having this discussion, it made me think how users will use these functions and what input items are relevant. Maybe I'll add a "holidays" function as well, because developers may need to check what's on the shelve in order to determine which ones to add/remove. They can also list holidays as an output if needed. So you'll see more changes coming 😀

purplesky2016:
- holidays module v0.13 fixed the bug I documented in the .py file, so I removed that note.

- Removed `year `component of the `add_holidays` keys, and let the function attach it. This way the developers don't have to maintain the input.

BryceStevenWilley (approved):
Looked over the new changes, and they all look great to me! Tested it as well, and it works! Had a small hiccup in the playground, since this doesn't work with `holidays=0.10`, but that shouldn't be an issue if you install it normally, since it'll install `holidays=0.13`.

PR #47, opened 2022-03-01
Allow an empty date field to be valid when the answer is not required

### LINKED ISSUE #44: BirthDate datatype doesn't respect `required: False`
Leaving the birthdate field empty doesn't let you continue right now. 
image

We need to figure out how to adjust the validation to take `required: False` into account. Might be possible with the new features allowed in Docassemble custom datatypes.

plocket:
Closes #44.

Sorry that GitHub is doing such a bad job of showing when changes are just whitespace changes. It was just easier for me to read with some indentation tweaks.

I would appreciate some testing on this. To test this, you'll have to edit the files and change the customdatatype names in the .py and the .yml because they could otherwise be overridden by our current customdatatypes. For me, they were overridden. I dealt with it by adding "Foo" before the names.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2021 AssemblyLine
+Copyright (c) 2022 AssemblyLine
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


nonprofittechy (approved):
Looks good, thanks! Let's rely on Black autoformatter for whitespace, etc. changes. I just added it to this repo.

Issue #42, opened 2022-01-17
Create a function that makes a `template` into a block that can be clicked on to copy to clipboard


nonprofittechy:
This is different from copy_share_button because it's for a longer block of text that isn't editable.

purplesky2016:
Do you want a separate function or to expand the existing copy_share_button module? From user perspective it might be better to use the same function for both purposes.

purplesky2016:
I'll investigate.

PR #52, opened 2022-03-21
Expand display_template() function

purplesky2016:
(As requested on the call, changed button color to secondary and renamed the new feature input names, but keep the original name scrollable to avoid breaking the code.)

Expansion changes:

1. Fix AL issue [#400](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/issues/400#issuecomment-1050352061) - add a css rule to collapse_template()
2. Changed `collapse_template_demo `to use `include` and added a link to` display_template`
3. Expanded `copy_button.py` to include template input type, and added a link  to` display_template` in its demo
4. Expanded `display_template `to include `copy `and `collapse`. Moved the function out of `misc.py` to a separate .py as it's getting too long. This should not break the code since it uses a package file. Fix issue #42.
5. Added styles.css for altoobox demos. This will provide convenience for future demos.


docassemble/ALToolbox/data/questions/display_template_demo.yml
```diff
@@ -1,42 +1,97 @@
...
+id: intro
+continue button field: intro
 question: |
   Display a markdown template 
 subquestion: |
-  ####Use cases:
+  When it comes to display a markdown template on the screen, you can use the **display_template()** function for the following features. 
+    
+  1. **scrollable** (default=True) 
+      - The default is to display the template with a stroll bar. This is helpful if your markdown template content is long.
```
nonprofittechy, line 26:
```suggestion
      - The default is to display the template with a scroll bar. This is helpful if your markdown template content is long.
```

docassemble/ALToolbox/display_template.py
```diff
@@ -0,0 +1,39 @@
+import re
+from .copy_button import *
+
+def display_template(template, scrollable=True, collapse=False, copy=False, class_name=None):
```
nonprofittechy, line 4: These names look good to me--easier to read/use than `copiable` and `collapsible` and I think easy to understand. Highlighting in case @plocket  or @BryceStevenWilley have more thoughts.

docassemble/ALToolbox/misc.py
```diff
@@ -84,24 +84,6 @@ def collapse_template(template, classname=None):
 <a class="collapsed" data-bs-toggle="collapse" href="#{}" role="button" aria-expanded="false" aria-controls="collapseExample"><span class="pdcaretopen"><i class="fas fa-caret-down"></i></span><span class="pdcaretclosed"><i class="fas fa-caret-right"></i></span> {}</a>
 <div class="collapse" id="{}"><div class="card card-body{} pb-1">{}</div></div>\
 """.format(the_id, template.subject_as_html(trim=True), the_id, classname, template.content_as_html())
-
```
nonprofittechy, line 87: We didn't talk about it, but are we thinking this function is just too big to go in misc.py?

nonprofittechy (approved):
Looks good to me, will wait to here on others' thoughts regarding the parameter names and placement in its own module vs misc.py

Edit: we also need to resolve the merge conflict

docassemble/ALToolbox/data/questions/display_template_demo.yml
```diff
@@ -1,42 +1,97 @@
...
+question: |
+  Copy button examples
+fields:  
+  - html: <h5>Copy and scrollable</h5>
+  - note: |      
+      ${ display_template(my_template1, copy=True) }
+  - html: <h5>Copy and collapse but not scrollable</h5> 
+  - note: | 
+      _(Hover over the box to see the full template content. If this is annoying, keep the default value scrollable=True as shown in the very last example.)_<br><br>
+      ${ display_template(my_template1, scrollable=False, collapse=True, copy=True) }
```
BryceStevenWilley, line 74: I played around with things, trying to get the box to have the correct height to show all of the content without requiring a mouse over, and I think the issue is that `my_template1.content_as_html(...)` returns everything in a textarea (I can't find where that happens in docassemble), and it seems that textareas in Firefox aren't resizable with CSS. Changing the textarea to a div, and just using `height: auto` in the CSS for that element works great, without any sort of sudden changes. 

I couldn't find a better solution, just wanted to document what I had found.

BryceStevenWilley (approved):
LGTM! I tried for a while trying to get rid of that mouse-over resizing, but I couldn't come up with a better solution at all, we can always improve things later.

docassemble/ALToolbox/data/questions/display_template_demo.yml
purplesky2016: I know what you mean. I tried <div> as well, it would work for the height, but won't trigger the button click reaction, because <div> is not an input type of field. (sigh)

plocket: If we want to copy from a `div` instead of a `textarea`, it may be possible as long as we do it a different way: https://stackoverflow.com/a/48020189/14144258.  We can also, as you folks have said, make it an issue for the future and I may have time to try it out then.

docassemble/ALToolbox/copy_button.py
```diff
@@ -6,18 +6,39 @@ def copy_button_html (
...
+
   button_str = '<div class="al_copy">\n'
   if ( text_before != '' ):
     button_str += f'<span class="al_copy_description">{word( text_before )}</span>\n'
-  button_str += f'<input readonly class="al_copy_value" type="text" value="{ text_to_copy }">\n'
-  button_str += f'<button class="btn btn-primary al_copy_button" type="button">\n'
+
+  # Add textarea tag if copy_template_block is True, along with docassemble template block class names
+  if copy_template_block:
+    button_str += f'<textarea readonly class=" card card-body{style_class} bg-light pb-1 al_copy_value {scroll_class}" placeholder="inert help text here" {adjust_height}>{ text_to_copy }"</textarea>\n'
```
plocket, line 23: I haven't been following quite closely enough, so I'm not sure if I'm missing context. I think this is missing a space and the placeholder text seems like it's not useful here, so I'll make a suggestion

```suggestion
    button_str += f'<textarea readonly class="card card-body {style_class} bg-light pb-1 al_copy_value {scroll_class}" {adjust_height}>{ text_to_copy }"</textarea>\n'
```

A separate question: What would we expect `adjust_height` to be?

docassemble/ALToolbox/data/questions/display_template_demo.yml
plocket: I'm looking at the js code and I think I'm not understanding what the problem is. I thought it was something wrong with the copy to text, but there's something about the copy button itself? Maybe we can talk about it during deep dive today.

docassemble/ALToolbox/display_template.py
```diff
@@ -0,0 +1,35 @@
...
+from .copy_button import *
+
+def display_template(template, scrollable=True, collapse=False, copy=False, class_name='bg-light'):   
+  # 1. Initialize    
+  if scrollable:
+    scroll_class = 'scrollable-panel'
+    adjust_height = ''
+  else:
+    scroll_class = ''
+    adjust_height = f'onmouseover="this.style.height = (this.scrollHeight) + \'px\';"'
```
plocket, line 11: Do we know if this works on mobile devices? It seems unlikely to. Sorry if I'm going over old ground.

docassemble/ALToolbox/copy_button.py
plocket: I see lower down that `adjust_height` is a function.  I'm not a fan of Hungarian notation, but if we anticipate developers using this function directly I think that some indication that this is a function could be helpful.

docassemble/ALToolbox/data/static/display_template.css
```diff
@@ -3,3 +3,19 @@
...
+  display_template() for a specific scenario */  
+  overflow: hidden;
+}
+.al_copy_block {
+  display: block;  
+}
+.al_copy_button span {
+  display: inline-block;
+}
+.al_copy, .card {
```
plocket, line 19: Making this a rule for the `card` class in general seems a little broad. Do we really need it to be for everything with the `card` class? Or is it specific to a display template card?

docassemble/ALToolbox/display_template.py
```diff
@@ -0,0 +1,35 @@
...
+def display_template(template, scrollable=True, collapse=False, copy=False, class_name='bg-light'):   
+  # 1. Initialize    
+  if scrollable:
+    scroll_class = 'scrollable-panel'
+    adjust_height = ''
+  else:
+    scroll_class = ''
+    adjust_height = f'onmouseover="this.style.height = (this.scrollHeight) + \'px\';"'       
+
+  class_name = ' ' + class_name.strip()
```
plocket, line 13: It looks like we're passing this to other functions. If so, I think we're sticking with a "black box" model of programming. That is, this function "doesn't know" how the next function will use this text. We should just hand down the plain text and let the future worry about handling whitespace.

Alternatively, if we really want to handle whitespace here (which I'd love to discuss), we should do it with the other strings as well, like `adjust_height`.

docassemble/ALToolbox/data/questions/display_template_demo.yml
purplesky2016: It's both: copy and tooltip would both stop working.

docassemble/ALToolbox/display_template.py
```diff
@@ -0,0 +1,35 @@
...
+      return f'<a class="collapsed" data-bs-toggle="collapse" href="#{the_id}" role="button" aria-expanded="false" aria-controls="collapseExample"><span class="pdcaretopen"><i class="fas fa-caret-down"></i></span><span class="pdcaretclosed"><i class="fas fa-caret-right"></i></span> {template.subject_as_html(trim=True)}</a><div class="collapse" id="{the_id}">{text}</div>'
+
+    # 2.2 If not collapsible, simply return output from copy_button_html()
+    else:  
+      return text
+
+  # 3. If not copiable, generate the whole output 
+  else:    
+    if not collapse:     
+      return f'<div class="{scroll_class} card card-body{class_name} pb-1" id="{the_id}"><div class="panel-heading"><h3>{template.subject_as_html(trim=True)}</h3></div>{template.content_as_html()}</div>'
```
plocket, line 32: To match the suggestion above to handle whitespace at the point where it's needed.
```suggestion
      return f'<div class="{scroll_class} card card-body {class_name} pb-1" id="{the_id}"><div class="panel-heading"><h3>{template.subject_as_html(trim=True)}</h3></div>{template.content_as_html()}</div>'
```

docassemble/ALToolbox/display_template.py
```diff
@@ -0,0 +1,35 @@
...
+
+    else:   
+      return f'<a class="collapsed" data-bs-toggle="collapse" href="#{the_id}" role="button" aria-expanded="false" aria-controls="collapseExample"><span class="pdcaretopen"><i class="fas fa-caret-down"></i></span><span class="pdcaretclosed"><i class="fas fa-caret-right"></i></span> {template.subject_as_html(trim=True)}</a><div class="collapse" id="{the_id}"><div class="{scroll_class} card card-body{class_name} pb-1">{template.content_as_html()}</div></div>'
```
plocket, line 35: To match the suggestion above to handle whitespace at the point where it's needed.
```suggestion
      return f'<div class="{scroll_class} card card-body {class_name} pb-1" id="{the_id}"><div class="panel-heading"><h3>{template.subject_as_html(trim=True)}</h3></div>{template.content_as_html()}</div>'

    else:   
      return f'<a class="collapsed" data-bs-toggle="collapse" href="#{the_id}" role="button" aria-expanded="false" aria-controls="collapseExample"><span class="pdcaretopen"><i class="fas fa-caret-down"></i></span><span class="pdcaretclosed"><i class="fas fa-caret-right"></i></span>{template.subject_as_html(trim=True)}</a><div class="collapse" id="{the_id}"><div class="{scroll_class} card card-body {class_name} pb-1">{template.content_as_html()}</div></div>'
```

docassemble/ALToolbox/copy_button.py
```diff
@@ -6,18 +6,39 @@ def copy_button_html (
...
+    button_str += f'<button class="btn btn-secondary al_copy_button" type="button">\n'
+
+  # Add tooltip texts
   button_str += f'<span class="al_tooltip al_tooltip_inert">{word( tooltip_inert_text )}</span>\n'
   button_str += f'<span class="al_tooltip al_tooltip_active">{word( tooltip_copied_text )}</span>\n'
+
+  # Add icon and label for the button
   button_str += f'<i class="far fa-copy"></i>\n'
-  button_str += f'<span>{word( label )}</span>\n'
+  button_str += f'<span>{word( label )}</span></button>\n'
```
plocket, line 41: Great catch!

docassemble/ALToolbox/display_template.py
purplesky2016: I don't think so. The original copy button function uses similar mouse actions therefore not for mobile either. We would need a separate issue for that.

docassemble/ALToolbox/data/static/display_template.css
purplesky2016: Good point. I'll fix that.

docassemble/ALToolbox/data/static/styles.css
```diff
@@ -0,0 +1,10 @@
+/* This file applies to interviews in ALToolbox only
```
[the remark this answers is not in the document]
purplesky2016: There is a warning in the demo file:
`---
This css file is used for the demo only, do not include it in your interview
features:
  css: styles.css`

It's won't change any demo files unless you specifically include this css file in your demo.

nonprofittechy:
> The conflicts seem to be caused by the pfs/black formatting. Maybe we can just choose this code, ignoring the version from `main`, and then run pfs/black again?

Sure, that's fine. black will run on `main` automatically after we merge.

docassemble/ALToolbox/data/static/display_template.css
```diff
@@ -3,3 +3,19 @@
...
+/* For the expansion of the copy_button_html() function*/
+textarea {
+  width: 100%;
+  /*The following rule takes effect only with the inline onmouseover event as generated in 
+  display_template() for a specific scenario */  
+  overflow: hidden;
+}
+.al_copy_block {
+  display: block;  
+}
```
purplesky2016, line 15:
```suggestion
   margin-top: 0.5rem;
}
```

docassemble/ALToolbox/display_template.py
purplesky2016: Sorry @BryceStevenWilley, you didn't add line 13 above with blank space, my memory got mixed up. That line was carried over from `collapse_template()`

docassemble/ALToolbox/data/static/display_template.css
```diff
@@ -3,3 +3,19 @@
...
+.al_copy_button span {
+  display: inline-block;
+}
+.al_copy, .card {
+  margin-bottom: 0.5rem;
+}
```
purplesky2016, line 21:
```suggestion
```

docassemble/ALToolbox/data/static/display_template.css
```diff
@@ -12,10 +12,8 @@ textarea {
 }
 .al_copy_block {
   display: block;  
+   margin-top: 0.5rem;
```
purplesky2016, line 15:
```suggestion
  display: block;  
  margin-top: 0.5rem;

docassemble/ALToolbox/data/static/styles.css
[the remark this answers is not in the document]
plocket: Somehow didn't notice that, sorry 👍

docassemble/ALToolbox/display_template.py
plocket: Before I forget - where does this get changed back again? I can't find a `'onmouseout'`.

plocket: I tested it on my phone and it works just fine! When I tap on the text, the textarea expands to its full height.

BryceStevenWilley: It doesn't get changed back, because the idea is that it should be fully extended at the start, but it's not possible without JS. Good to hear that it works on mobile though.

plocket:
How hard would it be for you to double check `tabbed_templates_html()` in [docassemble/ALToolbox/misc.py](https://github.com/SuffolkLITLab/docassemble-ALToolbox/pull/52/commits/a93ee49a2f68d76fd92f510ab91d1d2671a76c6f#diff-034d5d43a15a1498bfac222b1794b29fab2b294e438f87536a2253eb4d916226) to make sure that it still works, @BryceStevenWilley? We did our best to just keep the code from `main`. Using a diff checker, all that was added was a newline that shouldn't affect anything. Maybe just a diff check from another human would be enough.

The code @purplesky2016 was working on was v0.3.7, so it was hard to tell.

docassemble/ALToolbox/data/questions/display_template_demo.yml
purplesky2016: I'll resolve this conversation and create a new issue for changing `textarea `to `div`

BryceStevenWilley:
`tabbed_templates_html()` looks fine to me! No difference that I can tell, either in the diff or in the working component.

Issue #51, opened 2022-03-09
Add a simple "checkbox" function


nonprofittechy:
The code to insert conditional checkbox into a DOCX template takes up a lot of space in the template.  A simple shortcut function, maybe like `docx_checkbox(bool)` would let you make more compact and readable DOCX templates.

It would insert `[  ]` if the `bool` were `False`, and `[X]` if it were `True`.

It could certainly be a Jinja2 macro as well, but making it a globally available function might be nicer. This is a tiny convenience but one that might get used dozens of times in a single template.

PR #56, opened 2022-04-04
add function output_checkbox()

purplesky2016:
Fix issue #51.


docassemble/ALToolbox/misc.py
```diff
@@ -174,3 +172,10 @@ def add_records(obj, labels):
         obj[index].reference = val[1]
         index += 1
     return obj
+
+def output_checkbox(value:bool=None):
+    """Generate a conditional checkbox for docx templates"""
+    if value:
+      return '[X]'
+    else:
+      return '[  ]'
```
nonprofittechy, line 181:
```suggestion
def output_checkbox(value_to_check:bool, checked_value:str="[X]", unchecked_value:str="[  ]"):
    """Generate a conditional checkbox for docx templates"""
    if value_to_check:
        return checked_value
    else:
        return unchecked_value
```

we shouldn't use the keyword `value` because it's already a Docassemble function.

nonprofittechy (approved):
Looks good to me as long as the suggested change is OK!

docassemble/ALToolbox/misc.py
purplesky2016: Right!
I'll add an example in docx:

````
Or you can override the default checkbox symbols (strings):

Is there fresh air? {{output_checkbox(fresh_air, checked_value=”(x)”, unchecked_value=”( )”) }}`
```

Issue #67, opened 2022-05-16
Add save_input_data module 


purplesky2016:
This module will have a function to save all the input answer data. The function will be called by a `survey type` interview, and the saved data will be used in `InterviewStats` to generate `data reports`.

The module will be referenced in AL in order to avoid dependency on `InterviewStats`.

PR #68, opened 2022-05-16
Fix issue #67

purplesky2016:


docassemble/ALToolbox/save_input_data.py
```diff
@@ -0,0 +1,52 @@
...
+        field_dict[k] = list(v)[0]  # Field name without quotes
+
+    data_to_save = dict()
+    data_to_save["title"] = title
+    data_to_save["field_type_list"] = type_dict  # Needed later in process_data
+
+    for k, v in type_dict.items():
+        # If a field is of checkboxes type, flatten its elements dict
+        # so that each key/value pair is saved in its own column.
+        if v in ["checkboxes", "multiselect"]:
```
nonprofittechy, line 35: Suggestion: we could eliminate the extra dictionary (storing the types) if we use a check like this: `isinstance(item, DADict)`

purplesky2016: Hmm. What about `data_to_save['field_type_list'] = type_dict` which is part of the data to be saved, and later used in the Stats? Or maybe I misunderstood what you meant?

nonprofittechy: @purplesky2016 check out https://github.com/SuffolkLITLab/docassemble-ALToolbox/pull/68/commits/4714fb07d18640f3d574979e3805ead4d289b592 when you have a chance--this is the same output without making the author specify the type

PR #72, opened 2022-06-27
Add income module. Close #49.

### LINKED ISSUE #49: Add income tool
* A specific need for representing jobs:
   * Employer
      * Hourly or salaried
         * Each of those has a frequency (period)
      * Commissions and bonuses, tips
         * Those might also happen more than once a year
         * Some might be pretty predictable and be a big part of your income, esp. Commissions
   * Thinking that the right way to represent a job like this would be as a dictionary (or possibly a list-might be most flexible to make it a list). Each key would be one kind of income with an associated period OR even a deduction:
      * Wages
      * Commission
      * Tips
      * Tax deductions
      * Other? 
         * That could accumulate multiple “other” values, like union dues, etc. that an interview author didn’t think of in advance
      * If someone uses this, they might be okay with separating each job into multiple screens
   * In a common situation someone might have 9 individual items to list for one job!
   * Another idea:
      * Help someone enter their last 3 paystubs and then average them for the user
* Where should changes be? (in the module, or should developer’s fork?)
   * Let’s copy the GBLS version (https://github.com/GBLS/docassemble-income/blob/master/docassemble/income/income.py) and put it in the ALToolbox

Deep dive video 02/28/2022. Starts discussion at 3:50. Discussion of agreed structure starts at ~18:15 (starts as object of periodic values, moves to list of periodic values at 29:06).

Substitute class for `JobList` and `Job`. `CompoundJob` or `MultipartJob` that has a list of objects that are basically `PeriodicValue`s/`Income`s - a period with an amount earned or deducted from that period - with more attributes. Maybe that are equivalent to the current `Job` class? Duck-typing an income list. Would return the amount (for a period? whatever's there now). Maybe use `SimpleValue` or a similar existing object to be a periodic income amount.

plocket:
Adds an income module, including the ability to collect info for a job that has multiple incomes and deductions. Meant for use in things like financial statements.

The demo is a bit of a mess, as I've said, but feel free to have a go at it. Does not yet implement:
- Pre-built financial statement YAMLs
- A way for itemized jobs to return a list of their items (though the demo does show how a developer can do that for themselves)


plocket:
Not sure why the setup file changed or if it needs to be fixed.

docassemble/ALToolbox/data/questions/income.yml
```diff
@@ -0,0 +1,22 @@
...
+    - "Weekly"
+  - 
+    - 24
+    - "Twice per month"
+  - 
+    - 26
+    - "Once every two weeks"
+  - 
+    - 4
+    - "Once every 3 months"
```
BryceStevenWilley, line 21: This should be the same thing in YAML, and it's a lot more concise and a bit less weird (the `-` with nothing else on the line is particularly off IMO)
```suggestion
  - [12, "Monthly"]
  - [1, "Yearly"]
  - [52, "Weekly"]
  - [24, "Twice per month"]
  - [26, "Once every two weeks"]
  - [4, "Once every 3 months"]
```

BryceStevenWilley (commented):
Very quick look through, but a few things would make it easier to review.

> Not sure why the setup file changed or if it needs to be fixed.

Looks like it just updated pandas, and added my name to the README, so I don't think it needs to be fixed.

docassemble/ALToolbox/income.py
```diff
@@ -0,0 +1,1081 @@
+# Based on https://github.com/GBLS/docassemble-income/blob/master/docassemble/income/income.py
+
+from docassemble.base.util import (
```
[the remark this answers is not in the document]
plocket: Weird. It's `al_income.py`, though. I'll have to delete the other one.

docassemble/ALToolbox/data/questions/income.yml
```diff
@@ -0,0 +1,10 @@
+---
+variable name: al_times_per_year_list
+data:
+  - [12, "Monthly"]
```
nonprofittechy, line 4: This is a short list so I recommend sorting by frequency. Eg., make "weekly" first and then ascending size order, with "yearly" last.

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+
+
+def al_flatten(listname, index=1):
+    """
+    Return just the nth item in an 2D list. Intended to use for multiple choice
+    option lists in Docassemble. e.g., al_flatten(al_asset_source_list()) will
+    return ['Savings','Certificate of Deposit'...].
+    """
+    return [item[index] for item in listname]
+
```
nonprofittechy, line 29: Is this function actually used? It's so trivial that I'm not sure we should keep it.

I also find the prefixing very distracting for the function names. It's not idomatic Python and it's not something we're doing with other toolbox modules.

docassemble/ALToolbox/data/questions/income.yml
```diff
@@ -0,0 +1,10 @@
+---
```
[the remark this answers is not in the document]
plocket: I'll include the .py. For some reason I though they were supposed to be independent of each other.

> Maybe this should also be named al_income.yml

Hehe, I noticed that too

nonprofittechy:
Stopping here for now, will circle back to finish review tomorrow hopefully. Overall, looking in great shape, most of the comments are about style or functions that you probably haven't had to use yet so might have had an unclear purpose.

Thanks for taking on this massive task!

docassemble/ALToolbox/al_income.py
plocket: I was trying to differentiate them from the GBLS module or other packages with broad names. I'm happy with whatever, so just let me know what you'd like.

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+        # If an item has its own period, use that instead
+        frequency_to_use = self.times_paid_per_year
+        if hasattr(item, "times_paid_per_year") and item.times_paid_per_year:
+            frequency_to_use = item.times_paid_per_year
+
+        is_hourly = False
+        # Override if item should be calculated hourly (like wages). Both the job
+        # and the item itself need to declare they should be calculated hourly
+        if self.is_hourly and hasattr(item, "is_hourly"):
+            is_hourly = item.is_hourly
```
nonprofittechy, line 735:
```suggestion
        is_hourly = self.is_hourly and hasattr(item, "is_hourly") and item.is_hourly
```

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+        kwarg: times_per_year {str | num} (Optional) Number of times per year you
+            want to calcualte. E.g, to express a weekly period, use 52. Default is 1.
+        """
+        if times_per_year == 0:
+            return Decimal(0)
+
+        # If an item has its own period, use that instead
+        frequency_to_use = self.times_paid_per_year
+        if hasattr(item, "times_paid_per_year") and item.times_paid_per_year:
+            frequency_to_use = item.times_paid_per_year
```
nonprofittechy, line 729:
```suggestion
        if hasattr(item, "times_paid_per_year") and item.times_paid_per_year:
            frequency_to_use = item.times_paid_per_year
        else:
            frequency_to_use = self.times_paid_per_year     
```

nonprofittechy: Same suggestion here to use `times_per_year` not `times_paid_per_year`

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+            return (
+                value * Decimal(self.hours_per_period) * Decimal(frequency_to_use)
+            ) / Decimal(times_per_year)
+        else:
+            return (value * Decimal(frequency_to_use)) / Decimal(times_per_year)
+
+    def total(self, times_per_year=1, source=None):
+        """
+        Alias for ALItemizedJob.gross_total to integrate with ALIncomeList math.
+        """
```
nonprofittechy, line 753:
```suggestion
        """
        The total of an ALItemizedJob is defined as equivalent to the `gross` total, without accounting for deductions.
        """
```

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+        return self.gross_total(times_per_year=times_per_year, source=source)
+
+    def gross_total(self, times_per_year=1, source=None):
+        """
+        Returns the sum of positive values (payments) for a given times_per_year.
```
nonprofittechy, line 758:
```suggestion
        Returns the sum of positive values (like wages on a paycheck) for the specified time period.
```

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+                    value, times_per_year=times_per_year
+                )
+        return total
+
+    def deduction_total(self, times_per_year=1, source=None):
+        """
+        Returns the sum of money going out divided by a pay times_per_year as a
+        postive value. You can filter the items by `source`. `source` can be a
+        string or a list. If you use sources from money coming in, they will be
+        ignored.
```
nonprofittechy, line 787:
```suggestion
        Returns the total amount of "outgoing" values (deductions on a paycheck) for the specified time period.
        The total will be a positive value.
                
        You can filter the items by `source`. `source` can be a
        string or a list. If you use sources from money coming in, they will be
        ignored.
```

nonprofittechy: I'm not sure what the "sources from money coming in" ignored refers to. If the source doesn't apply it's not used, right? Is it adding different context?

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+        if not hasattr(self, "to_add"):
+            self.initializeAttribute("to_add", ALItemizedValueDict)
+        # Money being taken out
+        if not hasattr(self, "to_subtract"):
+            self.initializeAttribute("to_subtract", ALItemizedValueDict)
+
+    def item_value_per_times_paid_per_year(self, item, times_per_year=1):
+        """
+        Given an item and an times_per_year, returns the value accumulated by the
+        item for that `times_per_year`.
```
nonprofittechy, line 712: I'm confused by this API. How is the `item` parameter created?

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+    #    for key in values_dict.true_values():
+    #      value = values_dict[key]
+    #      result[key] = {}
+    #      result[key]['value'] = value.value
+    #      # Q: Include defaults for all attributes?
+    #      if hasattr(value, 'is_hourly'):
+    #        result[key]['is_hourly'] = value.is_hourly
+    #      if hasattr(value, 'times_paid_per_year'):
+    #        result[key]['times_paid_per_year'] = value.times_paid_per_year
+    #    return result
```
nonprofittechy, line 924:
```suggestion
```

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+
+
+class ALItemizedJobList(DAList):
+    """
+    Represents a list of jobs that can have both payments and money out. This is
+    a less common way of reporting income.
```
nonprofittechy, line 930:
```suggestion
    Represents a list of jobs that can have both payments and money out. 
    For example: a paycheck that has both wages and deductions, such as health insurance, 
    taxes, etc. 
    
    This is a less common way of reporting income.
```

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+        super().init(*pargs, **kwargs)
+        if not hasattr(self, "source") or self.source is None:
+            self.source = "Jobs"
+        if self.object_type is None:
+            self.object_type = ALItemizedJob
+
+    def total(self, times_per_year=1, source=None):
+        """
+        Alias for ALItemizedJobList.gross_total to integrate with
+        ALIncomeList math.
```
nonprofittechy, line 943:
```suggestion
        The total of an ALItemizedJobList is defined as the gross total, not accounting for any deductions.
```

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+        for job in self.elements:
+            total += job.deduction_total(times_per_year=times_per_year, source=source)
+        return total
+
+    def net_total(self, times_per_year=1, source=None):
+        """
+        Returns the net of the list's jobs (money in minus money out) divided by
+        the times_per_year. You can filter the items by `source`. `source` can be a
+        string or a list.
+
```
nonprofittechy, line 996:
```suggestion
        Returns the net of the list's jobs (money in minus money out) divided by
        the times_per_year. 
        
        For example, the net would reflect wages on a paycheck minus any 
        deductions, such as for health insurance or taxes.
        
        You can filter the items by `source`. `source` can be a string or a list.
```

nonprofittechy (changes requested):
As I said above, overall this is looking in good shape. I have a lot of suggestions for clarity. Overall comments:

- Use one term to replace `period`, not two. I think `times_per_year` works better than `times_paid_per_year` as not all income or deductions reflects a payment of any kind.
- Be more clear about what "in" and "out" mean. While it's not always as simple as wages or deductions, there's no reason to be vague. Just use an example and explain that it can reflect broader concepts sometimes too. But wages and health insurance, etc. are the most common.
- Avoid magic behavior like relying on intrinsicName and the `value()` function to get information that is internal to Docassemble or the YAML file. Prefer instead to use named parameters.
- Consider removing functions and classes that don't have an immediate use in a current financial statement, like the ledger code and the to_json() method that were both for very specific past projects.

docassemble/ALToolbox/al_income.py
nonprofittechy: If this is for internal use only, maybe prefix with `_`?

docassemble/ALToolbox/al_income.py
nonprofittechy: If it's not used in a current demo just remove it. Don't prefix functions, only classes.

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+            if int(index) == int(row[0]):
+                return row[1].lower()
+        return (
+            docassemble.base.functions.nice_number(int(index), capitalize=True)
+            + " "
+            + docassemble.base.functions.word("times per year")
+        )
+    except:
+        return ""
+    return ""
```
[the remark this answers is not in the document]
nonprofittechy: I also think now that we made the al_times_per_year_list a `data` block in the YAML the datastructure can be accessed more directly by an author, the function is less important.

nonprofittechy:
One more thing we should consider but can defer is some typing. That might make it easier to review too actually, but I think I was able to follow the logic as is without typing so far.

docassemble/ALToolbox/al_income.py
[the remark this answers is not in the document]
plocket: Shall I just remove the function?

docassemble/ALToolbox/al_income.py
plocket: It's just a warning that there's already a filter on this list that excludes money coming in (payments, etc.)

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+
+    def total(self):
+        """
+        If desired, to use as a ledger, values can be signed. Setting
+        transaction_type = 'expense' makes the value negative. Use min=0 in that
+        case.
+
+        This can't be used in an ALIncomeList because its `total` can retrun
+        positive and negative values, which would mess up ALIncomeList math
+        to add up all positive item amounts.
```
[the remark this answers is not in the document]
plocket: I accepted this, but want to note that I initially wasn't sure what "signed" meant and I'm not sure I would have understood without my previous knowledge.

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+    """
+
+    def init(self, *pargs, **kwargs):
+        super().init(*pargs, **kwargs)
+
+    def __str__(self):
+        """Returns just the name of the object, instead of its whole path."""
+        name_end = re.sub(r"^.*\['", "", self.instanceName.split(".")[-1])
+        isolated_name = re.sub(r"']$", "", name_end)
+        return isolated_name
```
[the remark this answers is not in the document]
plocket: Maybe it's not necessary. I'll have to check again. Even given the background explanation below, maybe there's something like a `.name` var devs can use instead.

This was a function you suggested when I tried to display the name of the job dynamically in the title by just using the __str__ of the item. E.g. `${ myItem }`. They kept coming out as their complex variable name. If I recall correctly, I mostly used this for the dynamically added items that people added in for themselves, so I don't think they would need translation.

If they did need translation, maybe devs can use the name of the object as a key, just like the other translations.  They could have a separate question for dynamically added items where they use some other method.

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+
+    def init(self, *pargs, **kwargs):
+        super().init(*pargs, **kwargs)
+        self.object_type = ALVehicle
+
+
+class ALSimpleValue(DAObject):
+    """
+    Like a Value object, but no fiddling around with .exists attribute because
+    this is designed to be stored in a list, not a dictionary.
```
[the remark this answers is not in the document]
plocket: I've gone with "Represents a currency value." as I couldn't think of anything else to add other than implementation details.

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+            ("charity", "Church or charitable donations"),
+            ("loan payments", "Loan, credit, or lay-away payments"),
+            ("support", "Support to someone not in household"),
+            ("other", "Other"),
+        ]
+    )
+    return source_list
+
+
+class ALIncome(PeriodicValue):
```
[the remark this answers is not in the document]
plocket: What other methods are there? In the documentation, I only see are `.amount()`. The only other quirk I see is that the `currency()` function will return the value of the item, but I think that's an artifact of `.__str__()`. If not, I'm not sure how to affect it.

Maybe we want to avoid triggering the gathering of `.value` if `.exists` is false? I think that's a behavior from which we diverged.

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1015 @@
...
+                    result += Decimal(job.net_total(times_per_year=times_per_year))
+        else:
+            for job in self.elements:
+                if job.source == source:
+                    result += Decimal(job.net_total(times_per_year=times_per_year))
+        return result
+
+
+class ALAsset(ALIncome):
+    """Like ALIncome but the `value` attribute is optional."""
```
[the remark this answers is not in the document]
plocket: Does this mean an `ALAsset` should have a `.market_value` instead of a `.value` and use that in its `.total()` method? Do there need to be other changes too? For example, for the `.balance`?

docassemble/ALToolbox/al_income.py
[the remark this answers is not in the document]
nonprofittechy: You could add `(mixed positive and negative)` as an explanation, `signed` just felt more specific to me. When I read the original version I wasn't sure if it was clear that positive and negative meant the literal value of the floating point values, and not their semantic meaning.

docassemble/ALToolbox/al_income.py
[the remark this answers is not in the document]
nonprofittechy: Maybe a quick walk through would help. Not sure if I get what this is doing yet and if there's a different solution that might be simpler. But if it's not part of the public API of the class it's not a big deal for now.

docassemble/ALToolbox/al_income.py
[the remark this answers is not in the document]
nonprofittechy: An asset can have both a value, which represents income, and a market value, which represents the asset's current worth.

E.g., if you have a checking account with a $1,000 balance that earns 1% interest each year:

- value is 10.0
- times_per_year would be 1.0
- market_value would be 1000.0

A financial statement would want you to list the two things separately. Usually there are two separate limits for income and assets. E.g., for SSI you can earn up to ~ $800/month and still receive something, but you're only allowed to have up to $2,000 in assets.

The total() of an IncomeList reflects the total income, not the total value of the items. You need to report both income from wages and income earned as interest on your savings and checking accounts, no matter how small those figures might be. So these still need to be two different methods.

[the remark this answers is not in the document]
nonprofittechy: Some assets, such as houses and vehicles, also have an associated loan. that's what the `balance` attribute is for. E.g., consider the example for a house that was owned by a landlord:

- value of 1500.0 (rent income)
- times_per_year of 12.0 (rent is paid monthly)
- market_value of 100000.0 (the current estimated sales price)
- balance of 80000.0 (mortgage)

For most people, houses and vehicles don't earn any income so the value is likely to be zero, but sometimes they do earn income.

docassemble/ALToolbox/al_income.py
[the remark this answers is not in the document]
nonprofittechy: I don't think we use .exists as a pattern in any interviews that I built with the income module. The functionality of that is replaced with the .source attribute.

I'm okay with telling people to use `currency(some_value.total())` instead of just `currency(some_value)`. I didn't realize the `currency` function had that shortcut but I think it's not super discoverable.

We might be able to get similar functionality by defining a custom  `__float__` method.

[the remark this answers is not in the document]
plocket: Based on this and the discussion in Slack, I'm adding just a `__str__` method, since `__float__` was described as an anti pattern.

docassemble/ALToolbox/al_income.py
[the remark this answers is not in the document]
plocket: Thanks for the explanation. Since ALAsset itself only deals with `.value` and never refers to `.market_value`, I'm not sure what to document about it. To be honest, I'm not sure what to document about `.market_value` - just that ALAsset can have one, though only the ALAssetList does anything with it? Also, if we're documenting `.market_value`, do we also need to document `.balance`?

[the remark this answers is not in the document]
nonprofittechy: Yes, both should be noted as the intent of the ALAsset is to provide one or both of those attributes. You can't put them in the class definition because that interferes with the way Docassemble resolves undefined attributes, but they should be mentioned in the docstring.

BryceStevenWilley (approved):
The comments I made were fixed, so marking my review as Approve.

plocket:
The above commit tries to work on a lot of function descriptions. I struggled to come up with some explanations and wording (like for the `source` attribute), so I'd appreciate thoughts on those.

Also tried some `__str__` stuff, but not sure if it's really what we're looking for.

nonprofittechy:
If you're okay @plocket I have some minor nits that I can try to tackle--new ones that I didn't see before. Hate to keep delaying this so I can just add my own commit with a fix.

- num is not a Python type so weird to see in the docstring-- lets use `int` and `float` or `Decimal`
- very confused by times_per_year ever being a string you convert to a float or integer. Maybe you felt limited by the DA datatype, but keep in mind you can use `datatype: number` in combo with `input type: radio` etc.

I will take another look through at the same time. Again, heroic efforts on a very large feature.

nonprofittechy:
@plocket you should give my last 3 commits a quick sanity check, and if there's nothing that seems incorrect I'll go ahead and merge.

nonprofittechy (approved):
Looks good to me with latest changes--will wait for @plocket to comment same

docassemble/ALToolbox/al_income.py
```diff
@@ -271,7 +325,12 @@ def init(self, *pargs, **kwargs):
         super().init(*pargs, **kwargs)
         self.object_type = ALJob
 
-    def total(self, times_per_year=1, source=None):
+    def total(
+        self,
+        times_per_year: float = 1,
+        source: Union[List[str], str] = None,
+        owner: str = None,
```
plocket, line 332: Though `owner` is up here, it's not passed into the call of `self.gross_total()` in the function.

docassemble/ALToolbox/al_income.py
```diff
@@ -189,45 +236,47 @@ def total(self, times_per_year=1, source=None, owner=None):
...
+    .net {float} (Optional) The net that the job makes during its pay period -
         its annual frequency. E.g. if the annual frequency is 52 (weekly), net
         is the total amount made for one week.
     .employer {str} (Optional) A single string for an employer's full name
     .employer_address {str} (Optional) A single string for employer's address
     .employer_phone {str} (Optional) An employer's phone number
     """
 
-    def total(self, times_per_year=1, source=None):
+    def gross_total(self, times_per_year: float = 1) -> Decimal:
```
plocket, line 265: Looks like I totally got this wrong. I'm a bit confused about how `super` works, so I'm not sure if that needs to be used here with `self.total()`. That's the only question I have here, though.

docassemble/ALToolbox/al_income.py
```diff
@@ -338,24 +403,30 @@ def net_total(self, times_per_year=1, source=None):
...
-        the value. E.g. 12 for a monthly value.
-    .value {num} (Optional) Represents an amount of money accumulated by this
-        asset over its `times_per_year`. Default is 0.
+    An ALAsset represents an asset that has a market value, an optional income
+    that the asset earns, and an optional balance which may be helpful if the
+    asset represents a financial account rather than a physical asset.
+
+    Can be stored in an ALAssetList.
+
+    Attributes: .market_value {float | Decimal} Market value of the asset.
```
plocket, line 412: Is this missing a new line character?
```suggestion
    Attributes:
     .market_value {float | Decimal} Market value of the asset.
```

docassemble/ALToolbox/al_income.py
```diff
@@ -441,19 +513,19 @@ def owners(self, source=None):
...
-    .balance {num} Balance of an asset.
-    .times_per_year {str | num} Represents a number of the annual frequency of
-        the value. E.g. 12 for a monthly value.
-    .value {num} (Optional) Represents an amount of money accumulated by this
-        asset over its `times_per_year`. Default is 0.
+    .year {str} The model year of the vehicle, like 2022
+    .make {str} The make of the vehicle, like "Honda"
+    .model {str} The model of the vehicle, like "Accord"
+    .market_value {int | float} Market value of an asset.
+    .balance {int | float} Balance of an asset.
```
plocket: Considering your description of a balance further up, does a vehicle need a balance?

docassemble/ALToolbox/al_income.py
```diff
@@ -338,24 +403,30 @@ def net_total(self, times_per_year=1, source=None):
...
+
+    Attributes: .market_value {float | Decimal} Market value of the asset.
+    .balance {float | Decimal } Current balance of the account, e.g., like
+        the balance in a checking account, but could also represent a loan
+        amount.
+    .value {float | Decimal} (Optional) Represents the income the asset earns
+        for a given `times_per_year`, such as interest earned in a checking
+        account. If not defined, the income will be set to 0, to simplify
+        representing the many common assets that do not earn any income.
+    .times_per_year {float} Number of times per year the asset earns the
```
plocket, line 421: Is `times_per_year` also optional if `value` is optional?

docassemble/ALToolbox/al_income.py
```diff
@@ -441,19 +513,19 @@ def owners(self, source=None):
...
-        the value. E.g. 12 for a monthly value.
-    .value {num} (Optional) Represents an amount of money accumulated by this
-        asset over its `times_per_year`. Default is 0.
+    .year {str} The model year of the vehicle, like 2022
+    .make {str} The make of the vehicle, like "Honda"
+    .model {str} The model of the vehicle, like "Accord"
+    .market_value {int | float} Market value of an asset.
+    .balance {int | float} Balance of an asset.
+    .value {float | Decimal} (Optional) Income earned by the vehicle (typically 0)
+    .times_per_year {str | float} Time frequency over which the `value` is earned
```
plocket: Is `times_per_year` also optional if `value` is optional? Also, is this still supposed to be a `{str | float}`? From what I can see elsewhere, only kwargs have that expected type.

docassemble/ALToolbox/al_income.py
```diff
@@ -555,36 +627,41 @@ def total(self, source=None):
...
-    .times_per_year {str | num} A number (or string representing a number)
-        representing the annual frequency of the job. E.g. 12 for monthly.
+    .times_per_year { float} A denominator of a year representing the annual
+         frequency of the job. E.g. 12 for monthly.
+    .exists {bool} (Optional) Allows an interview author to pre-define some common
+        descriptors, like "wages" or "union dues" without requiring the user to
+        provide a value for each item.
+
+        If the ".exists" attribute is False or undefined, the item will not be used
+        when calculating totals.
```
plocket, line 650: I think I may have just forgotten, but I don't clearly remember working with `.exist` for this class. Maybe it's a docassemble thing. Either way, I just wanted to let you know and am willing to go with whatever makes sense.

docassemble/ALToolbox/al_income.py
```diff
@@ -708,33 +805,39 @@ def _item_value_per_times_per_year(self, item, times_per_year=1):
         is_hourly = self.is_hourly and hasattr(item, "is_hourly") and item.is_hourly
 
         # If an item's value doesn't exist, use a value of 0
+        # TODO: is this behavior correct, or should it force gathering the value?
+        # What does a no-value item in the list represent?
```
plocket, line 810: I _think_ it represented an item whose value wasn't gathered. Maybe it was an experience with a `show if` or something. Maybe it was from my lack of understanding of the context.

docassemble/ALToolbox/al_income.py
```diff
@@ -708,33 +805,39 @@ def _item_value_per_times_per_year(self, item, times_per_year=1):
...
+        self, times_per_year: float = 1, source: Union[List[str], str] = None
+    ) -> Decimal:
         """
         Returns the sum of positive values (payments) for a given times_per_year.
         You can filter the items by `source`. `source` can be a string or a list.
         If you use sources from deductions, they will be ignored.
 
         Args:
-        kwarg: times_per_year {str | num} (Optional) Number of times per year you
+        kwarg: times_per_year {str | float} (Optional) Number of times per year you
```
plocket: Sometimes this appears as `{float}` and sometimes as `{str | float}`. If that's intentional, it's fine with me.

docassemble/ALToolbox/al_income.py
nonprofittechy: `super()` is only needed if we want to use the parent class and the child class overrides a method by the same name.

docassemble/ALToolbox/al_income.py
nonprofittechy: `mypy` complained that we inherited the `total()` method from ALIncomeList but didn't use the same method signature. The `owner` attribute is only used for error-free typing. I agree it's a bit weird for this method.

plocket:
Going to make a review with a bunch of comments. Sorry to pepper them out one at a time before.

docassemble/ALToolbox/al_income.py
```diff
@@ -0,0 +1,1063 @@
...
+
+    Vehicles have a .year_make_model() method which facilitates
+    listing vehicles on many financial statement forms.
+
+    Attributes:
+    .year {str} The model year of the vehicle, like 2022
+    .make {str} The make of the vehicle, like "Honda"
+    .model {str} The model of the vehicle, like "Accord"
+    .market_value {int | float} Market value of an asset.
+    .balance {int | float} Balance of an asset.
```
nonprofittechy:
```suggestion
    .balance {int | float} Balance of an asset, which may represent the loan amount for a vehicle
```

plocket (commented):
These look like some amazing changes! I had some questions, but whatever you want to do with the comments is fine with me. I'm going to take the changes to the interview on faith - you know the context much better than me.

docassemble/ALToolbox/al_income.py
nonprofittechy: I fixed in a later commit already

docassemble/ALToolbox/al_income.py
nonprofittechy: Sorry, I had fixed that in the final commit--I know it's tricky to view a diff spanning multiple commits in a PR

docassemble/ALToolbox/al_income.py
nonprofittechy: I think this is also behavior we can adjust after trying on @CaroRob's form

PR #80, opened 2022-07-13
Add nice_county_name method

nonprofittechy:
This function's purpose is simply to remove the word "County" from a string. Google Autocomplete adds the word County but pleadings are often written in way that makes the word "County" redundant, especially if a user can sometimes type the county name longhand. Users will almost never type out something like "Suffolk County" by hand.


plocket (approved):
Looks good 👍 The only improvement I could see is adding a demo of it. I'll make a PR with it for this branch and you can choose what to do with it.

PR #86, opened 2022-07-21
Review widget

BryceStevenWilley:
A widget that allows for people to, in the middle of an interview without triggering a page reload, give reactions (thumbs up and thumbs down for now) and text feedback. A much more light weight solution than an entirely separate interview, that can help us triage which interviews need more attention, directly from our users. Screenshots at the bottom of this PR description.

To get the widget, you call the `reaction_widget` function, passing in:
* an `up_action` that will be executed on the server if the thumbs up is pressed,
* a `down_action` (same, but for the thumbs down)
* (optionally) a `feedback_action`, which will can get the actual feedback text with `action_argument('feedback_text')`.
Once you press either of the thumbs, a text input screen with a submit button appears, and once the text feedback is submitted, those fade away, leaving a final "thank you" message.
The user displayed text is also customizable.

I followed the existing pattern that the [al_document](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/2f2b55ad9f17a4c271bec35c03beae2f1e12f37d/docassemble/AssemblyLine/al_document.py#L1378-L1405) uses currently, it's pretty clean and flexible enough.

Despite mentioning otherwise in https://github.com/SuffolkLITLab/docassemble-GithubFeedbackForm/issues/17, it actually does make sense to add a feedback widget to ALToolbox, and just leave the idea of saving the feedback to the Feedback package.

Only other changes that were made in this PR were to allow passing in 'md' to the `fa_icon` function, which broke the icons before.

Open to CSS suggestions, I know I'm not the person to polish it to a tee so I just got something that looked okay to me, and didn't try to make it too flashy.

### The Thumbs widget (everything in and including the box)
Screenshot from 2022-07-21 15-56-47

### The Text portion of the feedback
Screenshot from 2022-07-21 15-56-56

### The final thanks
Screenshot from 2022-07-21 15-57-09


nonprofittechy (approved):
Looks like it works great! I think we can improve the styles a bit--using a lighter weight would be helpful, a little more padding around the box.

I also think we might decide to use a built-in Bootstrap component, like a modal, to simplify styling and make it easier to keep consistent with a Bootstrap theme: https://getbootstrap.com/docs/5.0/components/modal/

But I think it is modular enough that we can substitute that part later.

BryceStevenWilley:
> using a lighter weight would be helpful

What do you mean by this again? Like a lighter color (or just an outline) for the icons?

> I also think we might decide to use a built-in Bootstrap component, like a modal, to simplify styling and make it easier to keep consistent with a Bootstrap theme: https://getbootstrap.com/docs/5.0/components/modal/
Maybe even better: https://getbootstrap.com/docs/5.0/components/offcanvas/ (the one that comes up on the bottom)

That's fair. While those do look better, personally I have a visceral negative reaction to little popups like that on other websites, reminds me too much of stuff like https://how-i-experience-web-today.com/detail.html. If we want to stick with bootstrap, I like the idea of using a card (this version looks good: https://getbootstrap.com/docs/5.0/components/card/#titles-text-and-links).

BryceStevenWilley:
Using the bootstrap card looks like it solves most of the box styling problems. Would love to still get a review on if my JS and CSS is up to par as well.

 Here's what the widget looks like now:
Screenshot from 2022-07-22 11-46-25
Screenshot from 2022-07-22 11-45-09

nonprofittechy:
I like the card--very nice and clean look with that change. Looks ready to merge--will let you do that.

BryceStevenWilley:
After some conversation on today's standup, I changed the main keyword feature to be "review", not "reaction", and also changed some of the default wording in the widget. Should be ready to go, I'm going to merge

docassemble/ALToolbox/misc.py
```diff
@@ -179,6 +186,66 @@ def tabbed_templates_html(tab_group_name: str, *pargs) -> str:
...
+<div class="card" style="width: 20rem;">
+  <div class="card-body">
+    <p class="al-thumbs-widget">{word(thumbs_display)}</p>
+    <a href="{js_thumbs_up}" id="al-thumbs-widget-up"
+        class="btn btn-md btn-info al-thumbs-widget" aria-label="{word('Thumbs up')}">{fa_icon('thumbs-up', size='md')}</a>
+    <a href="{js_thumbs_down}" id="al-thumbs-widget-down"
+        class="btn btn-md btn-info al-thumbs-widget" aria-label="{word('Thumbs down')}">{fa_icon('thumbs-down', size='md')}</a>
+    """
+    if review_action:
+        review_area_id = review_action + "_area_id"
```
plocket, line 231: How likely is it that two review buttons could have the same `review_action` name? If that happens, this id might be duplicated on the page, which would result in the wrong text being sent for some of the elements.

docassemble/ALToolbox/data/static/review_widget.js
```diff
@@ -0,0 +1,56 @@
...
+    for (let new_elem of all_done_stage) {
+      $(new_elem).fadeIn(al_fade_speed);
+    }
+  }, al_fade_speed);
+}
+
+function altoolbox_thumbs_up_send(event_name, has_feedback) {
+  action_call(event_name);
+  after_click();
+  if (!has_feedback) {
```
plocket, line 34: The argument handed in for this is either 'true' or 'false'. `!has_feedback` checks for a truthy value. If the value was 'false' - a string, which is a truthy value, which would create the opposite behavior than what is desired.

docassemble/ALToolbox/data/static/review_widget.js
```diff
@@ -0,0 +1,56 @@
...
+      $(new_elem).fadeIn(al_fade_speed);
+    }
+  }, al_fade_speed);
+}
+
+function after_send() {
+  var feedback_stage = document.getElementsByClassName('al-review-text');
+  var all_done_stage = document.getElementsByClassName('al-after-review');
+
+  for (let elem of feedback_stage) {
```
plocket, line 21: If there are indeed multiples of these elements, would a click on one send element hide all the other elements too?

docassemble/ALToolbox/data/static/review_widget.js
```diff
@@ -0,0 +1,56 @@
+var al_fade_speed = 300;
+
+function after_click() {
+  var thumbs_stage = document.getElementsByClassName('al-thumbs-widget');
+  var feedback_stage = document.getElementsByClassName('al-review-text');
+
+  for (let elem of thumbs_stage) {
```
plocket, line 7: What versions of ie are we supporting? This uses `let`, which is generally good practice for `for`s, but is only supported from ie 11 onwards. If that's where we're at, the practice is generally used everywhere in the code.

Just as a point of information, I believe `var` wouldn't run into any problems for the code inside the `for` here.

docassemble/ALToolbox/misc.py
```diff
@@ -179,6 +186,66 @@ def tabbed_templates_html(tab_group_name: str, *pargs) -> str:
...
+        js_review = (
+            f"javascript:altoolbox_review_send('{review_action}', '{review_area_id}')"
+        )
+        widget += f"""
+      <p class="al-review-text al-hidden">{word(review_display)}</p>
```
plocket, line 236: For accessibility, should this (and similar elements) also have `aria-hidden` and then handle that attribute in the js code?

docassemble/ALToolbox/misc.py
BryceStevenWilley: This widget isn't designed to be used twice on the same page,and it doesn't make much sense to use like that.

docassemble/ALToolbox/misc.py
BryceStevenWilley: `al-hidden` uses `display: none` which is equivalent to `aria-hidden`, and the element doesn't show up in the accessibility tree.

docassemble/ALToolbox/data/static/review_widget.js
BryceStevenWilley: It's confusing, but the values in _python_ are either `'true'` or `'false'`.  Copying the line here:

```python
 js_thumbs_up = f"javascript:altoolbox_thumbs_up_send('{up_action}', {'true' if review_action else 'false'})"
 ```
 
 `'true'` and `'false'` are within the curly braces (`{`), which evaluate as python expressions, so it gets added in the surrounding as the exact string and not the python capitalized `True` and `False` (which is why I think it's like that in the first place).

docassemble/ALToolbox/data/static/review_widget.js
BryceStevenWilley: Very good point, probably should have used `var`. TBF I'm not sure if everything else we do / docassemble / bootstrap does supports IE < 11, but it would be silly if everything else is fine but not this.

docassemble/ALToolbox/data/static/review_widget.js
BryceStevenWilley: Nope, not designed for being used more than once on the same screen, I don't really see a use case there.

PR #90, opened 2022-08-18
Business day normalize date strings

nonprofittechy:
The existing code didn't seem to work with DADateTime objects because it was expecting dates formatted like `%m/%d/%Y`. Switched to using ISO date formats throughout and simplified the logic a bit. I'm not sure if the `non_business_days()` function is strictly needed but I left that basically alone for now as it's not necessary to the core functionality.

A simple test YAML to explore:

```yaml
modules:
  - .business_days
---
question: |
  What date do you want to check?
fields:
  - Date: the_date
    datatype: date
---
mandatory: True
question: |
  % if is_business_day(the_date):
  ${ the_date } is a business day
  % else:
  ${ the_date } is NOT a business day
  
  The next business day is ${ get_next_business_day(the_date) }
  % endif
```  

@purplesky2016 feel free to ignore but figured this was your baby and I realized I don't know when your summer break is over!

### DIFF
--- a/docassemble/ALToolbox/business_days.py
+++ b/docassemble/ALToolbox/business_days.py
@@ -3,7 +3,7 @@ import pandas as pd
 import datetime
 from datetime import date as dt
 from docassemble.base.util import as_datetime, DADateTime
-from typing import Union
+from typing import Union, Dict
 
 """
   External docs: 
@@ -15,7 +15,19 @@ from typing import Union
 
 def standard_holidays(
     year, country="US", subdiv="MA", add_holidays=None, remove_holidays=None
-) -> dict:
+) -> Dict[str, str]:
+    """
+    Get all holidays in the specified year, country, and state (or other subdivision).
+    Note that this draws on the "holidays" package which may deviate slightly from
+    holidays observed by a local court, but should be very close to accurate.
+
+    Returns a dictionary like:
+    {
+        "2021-01-01": "New Year's Day",
+        ...
+        "2021-12-25": "Christmas Day",
+    }
+    """
     # 1. Get standard holidays from python's holidays module
     countr_holidays = []
     countr_holidays = holidays.country_holidays(
@@ -49,9 +61,10 @@ def non_business_days(
     first_n_dates=0,
     last_n_dates=0,
 ) -> dict:
+    # TODO: this function may not be necessary, but check with @purplesky2016 before removing
     # 1. Collect weekends and standard holidays
     # 1.1 Get all saturdays and sundays in the given year
-    # Must use .strftime('%m/%d/%Y')to make it a string, otherwise will get 'TypeError'
+
     sundays = (
         pd.date_range(start=str(year), end=str(year + 1), freq="W-SUN")
         .strftime("%Y-%m-%d")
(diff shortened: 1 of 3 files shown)


docassemble/ALToolbox/business_days.py
```diff
@@ -53,12 +54,12 @@ def non_business_days(
     # Must use .strftime('%m/%d/%Y')to make it a string, otherwise will get 'TypeError'
```
BryceStevenWilley, line 54: I'd remove this comment or change it

docassemble/ALToolbox/business_days.py
```diff
@@ -120,64 +121,52 @@ def non_business_days(
...
-            remove_holidays=remove_holidays,
-        ).keys()
+    if not isinstance(date, DADateTime):
+        date = as_datetime(date)
+    if date.dow in [
+        6,
+        7,
+    ]:  # Docassemble codes Saturday and Sunday as 6 and 7 respectively
+        return False
+    if date.format("yyyy-MM-dd") in standard_holidays(
```
BryceStevenWilley, line 137: `standard_holidays` returns date objects, not strings, so the string will never be in the list of holidays. You'd have to change back to `non_business_days` to get it working again.

BryceStevenWilley (commented):
Not sure `is_business_days` works as expected

docassemble/ALToolbox/business_days.py
nonprofittechy: standard_holidays returns a dictionary, where the key is an ISO formatted date string.

BryceStevenWilley: Hm, that's a weird class. [This function](https://github.com/dr-prodigy/python-holidays/blob/122387489d4a7a98b58d873c4412f58f0e0f1469/holidays/holiday_base.py#L262-L289) is the critical function, it stores everything internally as a datetime obj, and transforms strings to the comparable datetime when using it. I was getting really confused yesterday when I was trying it in the REPL (see below). I guess my tests were wrong somehow when I thought it was broken.

TBH I'd keep the type as `dict`, and either remove the doc string you added, or change it to say it uses either strings or date objects as keys.

```
>>> from docassemble.ALToolbox.business_days import standard_holidays
>>> standard_holidays(2022)
{datetime.date(2022, 1, 1): "New Year's Day", datetime.date(2022, 1, 17): 'Martin Luther King Jr. Day', datetime.date(2022, 2, 21): "Washington's Birthday", datetime.date(2022, 4, 18): "Patriots' Day", datetime.date(2022, 5, 30): 'Memorial Day', datetime.date(2022, 6, 19): 'Juneteenth National Independence Day', datetime.date(2022, 6, 20): 'Juneteenth National Independence Day (Observed)', datetime.date(2022, 7, 4): 'Independence Day', datetime.date(2022, 9, 5): 'Labor Day', datetime.date(2022, 10, 10): 'Columbus Day', datetime.date(2022, 11, 11): 'Veterans Day', datetime.date(2022, 11, 24): 'Thanksgiving', datetime.date(2022, 12, 25): 'Christmas Day', datetime.date(2022, 12, 26): 'Christmas Day (Observed)'}
>>> 
```

PR #101, opened 2022-09-01
Fixed bugs in ALIncome:

### LINKED ISSUE #99: Round floats before turning them into decimals
Floats will be slightly off when being cast as decimals, which will cause issues in our unit tests and if users of the library don't use the currency function. We should truncate the value of the decimal when calculating totals.

Full conversation in [Teams](https://teams.microsoft.com/l/message/19:910fed53a41c433091cce11066c866ae@thread.tacv2/1662051213604?tenantId=78733fa9-540e-4eb8-bf29-73c4eeb63412&groupId=eaa9bd9d-cf39-4686-8f30-e55aa9d98c75&parentMessageId=1662051213604&teamName=Legal%20Innovation%20and%20Technology%20Lab&channelName=assembly-line-backend-development&createdTime=1662051213604&allowXTenantAccess=false).

BryceStevenWilley:
* Decimal floating point rep errors (fixes #99)
  Get the locale's `frac_digits`, which is the number of fractional digits in currency, and always rounds to it when using floats. This is safe if being used on inputs from docassemble's `currency` datatype, which truncates the user typed value to that number of digits (but still stores it in a float).
  Some notable examples of locales that don't use 2 digits of fractional currency (from [Locale Helper](https://lh.2xlibre.net/values/frac_digits/)):
  * all arabic locales use 3 digits
  * Japan, Korea, and Iceland use 0 digits (i.e. all round numbers)
* `total()` in itemized job not working:
  Was iterating over the dictionary itself, which just gives the keys (i.e. the sources / names), not the item itself. Didn't error because we're checking `hasattr(item, 'value')` before doing much else, assuming that the given value wasn't a string. Fixes by iterating over `items()` instead.

* Started running `mypy` and `unittest` in github actions, to ensure tests are run on each commit.
  * fixed a few other mypy issues still in the repository
  * adjusted workflow script to properly set the right locale (which we depend on being set to 2 for the unittests).

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -18,6 +18,7 @@ jobs:
       - name: locales
         run: |
             sudo locale-gen en_US.UTF-8
+            # TODO(#104): run the tests while set to a different locale
             sudo update-locale LANG=en_US.UTF-8
       - name: Check
         uses: actions/checkout@v2
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -4,14 +4,21 @@ on:
   push:
   workflow_dispatch:
 
+env:
+  LANG: "en_US.UTF-8"
+
 jobs:
-  test-assemblyline:
+  test-altoolbox:
     runs-on: ubuntu-latest
     name: Run python only unit tests
     env:
       ISUNITTEST: true
     steps:
       - run: sudo apt-get update && sudo apt-get -y install libcurl4-openssl-dev build-essential python3-dev libldap2-dev libsasl2-dev slapd ldap-utils tox lcov valgrind libzbar0
+      - name: locales
+        run: |
+            sudo locale-gen en_US.UTF-8
+            sudo update-locale LANG=en_US.UTF-8
       - name: Check
         uses: actions/checkout@v2
       - run: pip install virtualenv
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,0 +1,22 @@
+name: Run python-only unit tests
+
+on:
+  push:
+  workflow_dispatch:
+
+jobs:
+  test-assemblyline:
+    runs-on: ubuntu-latest
+    name: Run python only unit tests
+    env:
+      ISUNITTEST: true
+    steps:
+      - run: sudo apt-get update && sudo apt-get -y install libcurl4-openssl-dev build-essential python3-dev libldap2-dev libsasl2-dev slapd ldap-utils tox lcov valgrind libzbar0
+      - name: Check
+        uses: actions/checkout@v2
+      - run: pip install virtualenv
+      - run: virtualenv -p python3.8 .venv
+      - run: .venv/bin/pip install -r docassemble/ALToolbox/requirements.txt
+      - run: .venv/bin/pip install --editable .
+      - run: .venv/bin/python3 -m mypy .
+      - run: .venv/bin/python3 -m unittest discover
--- a/docassemble/ALToolbox/al_income.py
+++ b/docassemble/ALToolbox/al_income.py
@@ -40,15 +40,15 @@ __all__ = [
 
 
 def _currency_float_to_decimal(value: Union[str, float]) -> Decimal:
-    """Given a float (that was set by a docassemble currency datatype, so truncated at 2 decimal places)
-    returns the correct decimal value, without floating point representation errors
+    """Given a float (that was set by a docassemble currency datatype, so
+    rounded to the nearest `fractional_digit` decimal places), returns the
+    exact decimal value, without floating point representation errors
     """
     if isinstance(value, float):
         # Print out the value of the float, rounded to the smallest allowable amount in the
         # locale currency, and use this value to make the exact Decimal value
         digits = get_locale("frac_digits")
-        format_str = "{:." + str(digits) + "f}"
-        return Decimal(format_str.format(value))
+        return Decimal(f"{value:.{digits}f}")
     else:
         return Decimal(value)
 
--- a/docassemble/ALToolbox/al_income.py
+++ b/docassemble/ALToolbox/al_income.py
@@ -865,7 +865,7 @@ class ALItemizedJob(DAObject):
         if source:
             # Make sure we're always working with a list of sources (names?)
             # Add up all money coming in from a source
-            sources = self.source_to_list(source=source)
+            sources = self.source_to_set(source=source)
             for key, value in self.to_add.elements.items():
                 if key in sources:
                     total += self._item_value_per_times_per_year(
@@ -903,7 +903,7 @@ class ALItemizedJob(DAObject):
         if source:
             # Make sure we're always working with a list of sources (names?)
             # Add up all money coming in from a source
-            sources = self.source_to_list(source=source)
+            sources = self.source_to_set(source=source)
             for key, value in self.to_subtract.elements.items():
                 if key in sources:
                     total += self._item_value_per_times_per_year(
@@ -916,7 +916,7 @@ class ALItemizedJob(DAObject):
                     self._item_value_per_times_per_year(
                         item, times_per_year=times_per_year
                     )
-                    for item in self.to_subtract.elements
+                    for key, item in self.to_subtract.elements.items()
                 )
             )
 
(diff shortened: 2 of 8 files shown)


docassemble/ALToolbox/al_income.py
```diff
@@ -127,21 +142,20 @@ def __str__(self) -> str:
...
-                Decimal(self.value)
-                * Decimal(self.hours_per_period)
-                * Decimal(self.times_per_year)
+                val * Decimal(self.hours_per_period) * Decimal(self.times_per_year)
             ) / Decimal(times_per_year)
         else:
-            return (Decimal(self.value) * Decimal(self.times_per_year)) / Decimal(
-                times_per_year
-            )
+            return (val * Decimal(self.times_per_year)) / Decimal(times_per_year)
```
plocket, line 158: Math isn't my strong suit, so I'm thinking out loud here. Could this still return a value that doesn't conform to the locale format? That said, would we even want this to always return a value that conforms to the locale format or do we want developers to use `currency` for that? Though `val` should be (and is given these changes) a locale-formatted value to get the user input correct, maybe the rest of the math here should return a more accurate fraction so that further math operations could be based on the most accurate values.

docassemble/ALToolbox/test_al_income.py
```diff
@@ -0,0 +1,153 @@
...
+    def test_simple_value(self):
+        value = ALSimpleValue(transaction_type="expense", value=5)
+        self.assertEqual(Decimal(-5), value.total())
+        value.transaction_type = "asset"
+        self.assertEqual(Decimal(5), value.total())
+
+        value_str = ALSimpleValue(value="5.3")
+        self.assertEqual(Decimal("5.3"), value_str.total())
+        value_neg_str = ALSimpleValue(transaction_type="expense", value="5.3")
+        self.assertEqual(Decimal("-5.3"), value_neg_str.total())
```
plocket, line 30: Are we trying to test that a `Decimal` value is returned. If not, I think this may not be as clear as it could be because it uses `Decimal()`. I'm ok with whatever choice you make, but if we wanted to we could do something like:
```py
self.assertEqual( "-5.3", str(value_neg_str.total()) )
```

It would be a pain, though.

docassemble/ALToolbox/test_al_income.py
```diff
@@ -0,0 +1,153 @@
...
+                "val1": ALItemizedValue(
+                    value=5.30, exists=True, is_hourly=False, times_per_year=1
+                ),
+                "val2": ALItemizedValue(
+                    value=5.30, exists=False, is_hourly=False, times_per_year=1
+                ),
+            }
+        )
+        itemized_dict.hook_after_gather()
+        self.assertEqual(1, len(itemized_dict))
```
plocket, line 127: Not relevant to this review: Would love to talk about this sometime and understand a little more.

docassemble/ALToolbox/test_al_income.py
```diff
@@ -0,0 +1,153 @@
...
+    ALAsset,
+    ALAssetList,
+    ALVehicle,
+    ALVehicleList,
+    ALSimpleValue,
+    ALSimpleValueList,
+    ALItemizedJob,
+    ALItemizedValueDict,
+    ALItemizedValue,
+    recent_years,
```
plocket, line 16: Is this used in here now? Or is it planned for future use?

docassemble/ALToolbox/al_income.py
```diff
@@ -38,6 +39,20 @@
 ]
 
 
+def _currency_float_to_decimal(value: Union[str, float]) -> Decimal:
+    """Given a float (that was set by a docassemble currency datatype, so truncated at 2 decimal places)
```
plocket, line 43: After a second look, I understand what you're saying here, but initially I thought you were saying, "We're accepting a value that's already been truncated and then we're truncating it." No suggestions at the moment, though. It's a very confusing concept.

plocket (approved):
This is a great addition and fixes some serious flaws in this code! I have a lot of comments, but I think most of them are due to my own lack of knowledge in this area. If I understand the purpose of this correctly, I think this is basically ready to go.

docassemble/ALToolbox/al_income.py
```diff
@@ -38,6 +39,20 @@
...
+    returns the correct decimal value, without floating point representation errors
+    """
+    if isinstance(value, float):
+        # Print out the value of the float, rounded to the smallest allowable amount in the
+        # locale currency, and use this value to make the exact Decimal value
+        digits = get_locale("frac_digits")
+        format_str = "{:." + str(digits) + "f}"
+        return Decimal(format_str.format(value))
```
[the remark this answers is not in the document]
plocket: Does this round somewhere? It must, to pass the tests.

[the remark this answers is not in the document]
nonprofittechy: It's using Python's format string to provide a string representation of the floating point value, with a fixed number of digits.

Decimal accepts a `str` as input.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    name: Run python only unit tests
+    env:
+      ISUNITTEST: true
+    steps:
+      - run: sudo apt-get update && sudo apt-get -y install libcurl4-openssl-dev build-essential python3-dev libldap2-dev libsasl2-dev slapd ldap-utils tox lcov valgrind libzbar0
+      - name: locales
+        run: |
+            sudo locale-gen en_US.UTF-8
+            sudo update-locale LANG=en_US.UTF-8
```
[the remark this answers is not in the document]
BryceStevenWilley: Made issue #104 for that, referenced in the code as well.

docassemble/ALToolbox/test_al_income.py
BryceStevenWilley: I'm not sure I understand why it's not clear because it uses Decimal: that's the only type we can use to make sure that the exact value is being preserved. If you were confused about why I'm passing a string to Decimal, it's because it's the easiest way to pass in an exact value.

However, I think you're right in that the thing we're ultimately trying to test is that the right value would be printed to the document. I'll add a few of those tests where possible, but will keep most of them testing `Decimal == Decimal`.

docassemble/ALToolbox/test_al_income.py
BryceStevenWilley: The quick gist is that I wanted to test the behavior of `hook_after_gather`, which removes  elements where `x.exists == False`. IMO, I'm not sure that's the best behavior of the function, but it should be tested at least.

docassemble/ALToolbox/test_al_income.py
BryceStevenWilley: Future use. Would like most of the things to be tested, but not sure how to make a tests that doesn't just duplicate the function itself.

docassemble/ALToolbox/al_income.py
BryceStevenWilley: As confusing as it is, it's kinda true? Docassemble does round (not truncate, I'll change that wording. I was using truncate to mean get rid of additional decimal places, not necessarily to mean `Math.floor` like it sometimes is) the value to the nearest hundredth place, but it's still storing the value in a floating point variable, which isn't exact. So we round it again, and store the rounded value in a variable that can represent it exactly.

docassemble/ALToolbox/al_income.py
BryceStevenWilley: Good catch! I had been assuming that users would use `currency()` for that. There are some ways that we could make it always return a value with exactly `fraction_digits` decimal places, but I'm not sure we'd want to do that? I'm unsure how exactly this sort of accounting math is done, but it would only really happen when you have non-standard times, i.e. if you get paid monthly, what would you say your weekly pay is? (that question doesn't make too much sense IMO)  Or, if you get paid bi-weekly, what is your monthly pay? Does it depends on the month (we're not being that detailed with this library to answer it that way), or do you average things out?

I'm leaning towards leavings things as accurate as possible, since the results of these functions could be used in other equations; we don't want to round too early. For now I added a test that uses `Decimal::quantize` (the same function we would use to address this) to check that at least when rounding the value is still correct.

BryceStevenWilley: Pinging @nonprofittechy, I'm not sure if you could answer any of these accounting questions.

nonprofittechy: Where this would come up would be in income qualification questions. A user won't generally be penalized for averaging it slightly differently (e.g., the 30.416 days vs a 28 day month) if the form asks for "monthly income". I've never seen a form ask for "weekly" income except to help the user report it accurately--they're likely to ask for monthly or annual income. Keep in mind that for most of our client populations income can vary quite a lot from month to month.

For the real-world context, often, the user reports the approximate value on the form and the organization might ask for the last 4 pay stubs + most recent tax return and do their own math, but they want the approximate answer on the form.

I would suggest we do the possible rounding as close to the end of the calculation as possible so our default is more precision.

PR #106, opened 2022-09-08
Source excludes and other Income improvements

BryceStevenWilley:
Lots of changes that I've been using to make the Affidavit of Indigency Supplement. Another PR incoming with the changes to the questions themselves.

* adds `exclude_source` param to most functions that take the `source` param. Will filter out the items in that list / dict / etc. to those that don't match the `exclude_source`. Useful for filling in "other" text boxes in PDFs, i.e. `jobs.deduction_total(exclude_source=list_of_inputs_in_pdf)`
* added more mypy types, which found a missing return in a function
* adds a "fields" method to `ALItemizedValue`, which shows a checkbox that sets `.exists`, and a currency input for `.value`
* add cast functions to `ALItemizedValue`: `currency()` casts to float, so `__float__` defines how it will behave when the object is passed to `currency()`, and our `thousands()` method uses a format string, so `__format__` defines how it will behave when passed there
* add `ALItemizedValueDict.total()` function. The itemized value dict is useful for gathering a list of user expenses (which might occur on a time scale, so aren't `ALIncome`s or `ALSimpleValues`).
* pass a param to `ALItemizedJob` that lets you define the employer as an arbitrary type (so we use `ALIndividual`)


docassemble/ALToolbox/al_income.py
```diff
@@ -661,20 +673,58 @@ class ALItemizedValue(DAObject):
         when calculating totals.
     """
 
+    def fields(self, use_exists=True) -> List[Dict[str, Any]]:
```
nonprofittechy, line 676: Took me a second to figure out this was returning a list for purposes of the YAML `fields` specifier.

In the Weaver, our pattern is `ask_` when what the function returns is a `fields` list for use in YAML--maybe a similar change to the name here would make sense?

docassemble/ALToolbox/al_income.py
```diff
@@ -661,20 +673,58 @@ class ALItemizedValue(DAObject):
...
+    def __str__(self) -> str:
         """Returns a string of the value of the item with two decimal places."""
         currency_str = "{:.2f}".format(self.value)
         return currency_str
 
+    def __float__(self) -> float:
+        if hasattr(self, "exists") and not self.exists:
+            return 0.0
+        else:
+            return float(self.value)
```
nonprofittechy, line 720: Yeah we had debated adding this before but I think you're right to do so.

docassemble/ALToolbox/al_income.py
```diff
@@ -661,20 +673,58 @@ class ALItemizedValue(DAObject):
...
+
+    def __int__(self) -> int:
+        return round(float(self))
+
```
nonprofittechy, line 724: `int()` typically truncates a float, and does not round it. I'm not sure this is the right choice

nonprofittechy (approved):
LGTM. The only thing that sticks out is `__int__` rounding instead of truncating, but looks likes some solid changes. Just LMK if the rounding is intentional and can be justified--if so we should document that

docassemble/ALToolbox/al_income.py
BryceStevenWilley: Are you talking about [user_ask_about_field](https://github.com/SuffolkLITLab/docassemble-ALWeaver/blob/e866c3ee0146638f63a9e2cee130ff533446f2ef/docassemble/ALWeaver/interview_generator.py#L455)? I couldn't find any other instances of `ask` in methods. I copied that method name from AssemblyLine general, like [ALAddress](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/71157d4cce4956e7350a8ebf971375f8ce76bd70/docassemble/AssemblyLine/al_general.py#L72) and [ALIndividual](https://github.com/SuffolkLITLab/docassemble-AssemblyLine/blob/71157d4cce4956e7350a8ebf971375f8ce76bd70/docassemble/AssemblyLine/al_general.py#L545).

I can document thing more for sure. Some clearer type aliases would probably help.

docassemble/ALToolbox/al_income.py
BryceStevenWilley: Really was necessary for the attachment block

docassemble/ALToolbox/al_income.py
BryceStevenWilley: That's fair. I added it because the `thousands()` function will cast to int before putting it in a f-string if `show_decimal == False`, and IMO thousands truncating isn't expected behavior. I think you're right that the solution is to change the behavior of `thousands()` instead of changing this, and maybe truncating is expected behavior there.

nonprofittechy: Maybe we should let the use choose in the `thousands` function. I think it may depend on instructions in the form. Truncating may be slightly more expected in situations where someone is displaying their income on a form, but offering a choice seems fine.

docassemble/ALToolbox/al_income.py
nonprofittechy: `income_fields()` would match the ALGeneral pattern, that works too. And yeah just a docstring would be fine also

docassemble/ALToolbox/al_income.py
BryceStevenWilley: Decided to just truncate instead. I wasn't using `thousands` with `show_decimals=False` anyway, shouldn't be over-complicating the function for things we aren't using in production.

PR #108, opened 2022-09-08
Move questions from demo to al_income

BryceStevenWilley:
* made `ALItemizedJob` use `.source`, not `.name`, to align with `ALJob`
* only major language change was to say "job title" as opposed to "name of job"
* Added revisit and table blocks from most types


docassemble/ALToolbox/data/questions/al_income.yml
```diff
@@ -89,4 +89,602 @@ data: !!omap
...
+    Enter info from your paystub for your job as a ${ x.source }
+fields:
+    - ${ word("What is your hourly pay for this job?") if x.is_hourly else word("What are your wages for this job?") }: x.to_add['full time'].value
+      datatype: currency
+      show if:
+        code: x.is_part_time is False
+    - ${ word("What is your hourly pay for this job?") if x.is_hourly else word("What are your wages for this job?") }: x.to_add['part time'].value
+      datatype: currency
+      show if:
+        code: x.is_part_time is True
```
nonprofittechy, line 181: I don't think we should use the `word()` function here--it's a totally different translation system from the XLSX one. Maybe use the `label: field:` variation so there is room to use a Mako `if` statement and this will go to the regular translation file?

docassemble/ALToolbox/data/questions/al_income.yml
```diff
@@ -89,4 +89,602 @@ data: !!omap
...
+  Tell us about your vehicle
+fields:
+  - Year: x.year
+    input type: combobox
+    code: |
+      recent_years()
+  - Make: x.make
+    hint: e.g, Ford, Honda
+  - Model: x.model
+    hint: e.g., Fusion, Civic
```
nonprofittechy, line 545: I know I added these `hints` but I think we should replace with the normal `help` feature--`hint` text is disfavored: https://www.nngroup.com/articles/form-design-placeholders/

docassemble/ALToolbox/data/questions/al_income.yml
```diff
@@ -89,4 +89,602 @@ data: !!omap
...
+question: |
+  How many vehicles do you have?
+fields:
+  - no label: x.target_number
+    datatype: integer
+---
+comment: |
+  Notes:
+  
+  * Do not try to use the `total()` method with an ALVehicle. ALVehicleList.
```
nonprofittechy, line 579: Not sure about this comment. the `.total()` method should return the total income from all vehicles, which is generally speaking 0 but it's not semantically wrong. Also both comments are weirdly placed--probably belong in the class definition or docstring

nonprofittechy: That said there a number of similar comments and it's fine to leave most of them in--just suggest removing this one incorrect one for now

docassemble/ALToolbox/data/questions/al_income.yml
```diff
@@ -89,4 +89,602 @@ data: !!omap
...
+depends on:
+  - user_language
+language: es
+variable name: job_items_names
+data:
+  full time: salario del el trabajo de tiempo completo
+  part time: salario del el trabajo de medio tiempo
+  tips: Propinas
+  insurance: Seguro
+  federal_taxes: Impuestos federales
```
nonprofittechy, line 689:
```suggestion
---
variable name: job_items_names
data:
  full time: Full time wages
  part time: Part time wages
  tips: Tips
  insurance: Insurance
  federal_taxes: Federal taxes
  state_taxes: State taxes
  social_security: Social security
  medicare: Medicare
```

I don't think I realized when I first worked on this that the `data` block gets put in the translation file. This will make it more flexible and prevent errors if the `user_language` is set to `pt` or another language exclusive of `en` and `es`

nonprofittechy (approved):
LGTM, this looks like it's pretty useful as-is. We may want to think about splitting it up eventually so there are multiple variations of pre-canned income questions for different types of documents, and we're not limited to lowest common denominator in this file.

We should also document this and put it on the https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/label_variables page

docassemble/ALToolbox/data/questions/al_income.yml
BryceStevenWilley: never realized that those were separate, thanks. Do you mean it should change to something like this?:
```yaml
  - label: |
      % if x.is_hourly:
      What is your hourly pay for this job?
      % else:
      What are your wages for this job?
      % endif
  ```

File: docassemble/ALToolbox/data/questions/al_income.yml (hunk 3 above)
Comment (BryceStevenWilley) [reply docassemble/ALToolbox/data/questions/al_income.yml:579]: I'll remove that one.

These comments were originally in the `subquestion` of the block, because they used to be in the demo, not a production interview; I didn't want to move them too far.

**Comment by BryceStevenWilley** [comment]
> We should also document this and put it on the https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/label_variables page

I like the idea, but I'm not sure what exactly go on that page. A lot of the questions here are generic variables that wouldn't quite be picked up by the Weaver yet. Should it be a new tab under the "Feature Documentation" sidebar?

**Comment by nonprofittechy** [comment]
> > We should also document this and put it on the https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/label_variables page
> 
> I like the idea, but I'm not sure what exactly go on that page. A lot of the questions here are generic variables that wouldn't quite be picked up by the Weaver yet. Should it be a new tab under the "Feature Documentation" sidebar?

Good point--but maybe we should create a standard object attribute, like ALIndividual.jobs, etc. so that we can document a specific attribute?

For now just a quick overview of the class as an item under "Feature documentation" would work too.

File: docassemble/ALToolbox/data/questions/al_income.yml [hunk 5]
```diff
@@ -89,4 +89,595 @@ data: !!omap
...
+  Tell us about your vehicle
+fields:
+  - Year: x.year
+    input type: combobox
+    code: |
+      recent_years()
+  - Make: x.make
+    help: For example, Ford, Honda
+  - Model: x.model
+    help: For examlpe, Fusion, Civic
```
nonprofittechy, line 557:
```suggestion
    help: For example, Fusion, Civic
```

Issue #36, opened 2021-12-07
Escape button disappears on small screens


nonprofittechy:

nonprofittechy:
Upstream problem - items added to the horizontal menu just disappear on "small" or "medium"? screens.

But what it just can't fit? We think it should go on the next line.

plocket:
Without js, I can't set `aria-hidden` for this new small-screen Escape button. That means that when the screen is wide, I believe that makes it non-compliant with accessibility standards. Thoughts?

plocket:
Using Bootstrap, even the smallest breakpoint (`sm`) isn't small enough for the smallest screens. We can change the width of the smallest breakpoint with SASS vars, but it may cause issues with other docassemble responsive content. So it seems like it may not be possible to get what we want from an upstream change.

Therefore we've decided to try to use our own js and keep the escape button in the nav bar. We'll need to provide a warning to our developers that they will have to keep the content in the nav bar small enough to account for the button. The feasibility depends partly on how Bootstrap moves its nav elements to the dropdown menu and whether our code will be triggered before or after the Bootstrap code. It'll work best if we go last.

PR #111, opened 2022-09-09
Address #36, escape button disappears on small screens

plocket:
Address #36 with js by using the element already in the nav bar and just moving it to a different location in the nav bar - one that won't disappear when the screen gets small. Adds a demo of the escape button.

I've not used the `data` specifier much and I'm not sure how to make it available to translators.


nonprofittechy (approved):
Elegant and simple--thanks! I'll get in the translatability fix before I merge

PR #116, opened 2022-09-21
Make icons configurable for `collapse_template()`

### LINKED ISSUE #115: New feature request for collapse feature - customize icon please
I would like to be able use  + and - signs for expanding and collapsing text, or just be able to pick icons for these actions.
Thanks.

nonprofittechy:
Fix #115

Adds two new keyword parameters to `collapse_template`: `expand_icon` and `collapse_icon` which are assumed to be fontawesome icons in the "fa-solid" category. 

Also updated `fa_icon()` to use new Fontawesome 5 `fa-solid` class explicitly instead of the deprecated `fa` class.

### DIFF
--- a/docassemble/ALToolbox/misc.py
+++ b/docassemble/ALToolbox/misc.py
@@ -87,7 +87,7 @@ def fa_icon(icon: str, color="primary", color_css=None, size="sm", fa_class="fa-
         )
     else:
         return (
-            '<i class="fa fa-'
+            f'<i class="{fa_class} fa-'
             + icon
             + size_str
             + '" style="color:var(--'
--- a/docassemble/ALToolbox/misc.py
+++ b/docassemble/ALToolbox/misc.py
@@ -141,10 +141,18 @@ def number_to_letter(n: Optional[int]) -> str:
 
 
 def collapse_template(
-    template, classname=None, expand_icon="caret-down", collapse_icon="caret-right"
+    template,
+    classname=None,
+    closed_icon="caret-right",
+    open_icon="caret-down",
 ):
     """
     Insert HTML for a Bootstrap "collapse" div.
+
+    Optionally, you can specify a custom icon to override the defaults:
+
+    The default icons are "🞂" which displays when the text is collapsed (`closed_icon`) and
+    "▼" which displays when the text is open (`open_icon`).
     """
     if not template.subject_as_html(trim=True) and not template.content_as_html():
         return ""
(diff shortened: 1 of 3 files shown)


docassemble/ALToolbox/misc.py
```diff
@@ -62,7 +62,9 @@ def tel(phone_number) -> str:
     return '<a href="tel:' + str(phone_number) + '">' + str(phone_number) + "</a>"
 
 
-def fa_icon(icon: str, color="primary", color_css=None, size="sm") -> str:
+def fa_icon(
+    icon: str, color="primary", color_css=None, size="sm", fa_class="fa-solid"
```
BryceStevenWilley, line 66: Would help to give some examples of `fa_class` in the doc string below.

docassemble/ALToolbox/misc.py
```diff
@@ -78,7 +80,7 @@ def fa_icon(icon: str, color="primary", color_css=None, size="sm") -> str:
         return ":" + icon + ":"  # Default to letting Docassemble handle it
     elif color_css:
         return (
-            '<i class="fa fa-'
+            f'<i class="{fa_class} fa-'
```
BryceStevenWilley, line 83: Is `fa-solid` the same as `fa`? I can't actually find if it says that in the docs, though `fa-solid` is listed first in https://fontawesome.com/v6/docs/apis/javascript/import-icons#referencing-icons.

docassemble/ALToolbox/misc.py
```diff
@@ -138,7 +140,9 @@ def number_to_letter(n: Optional[int]) -> str:
     return string
 
 
-def collapse_template(template, classname=None):
+def collapse_template(
+    template, classname=None, expand_icon="caret-down", collapse_icon="caret-right"
```
BryceStevenWilley, line 144: nit: I'd use `expanded_icon` and `collapsed_icon`, so it's clear that those are the icons being used in those states. Right now it's vague to me if expand should be the icon to click to expand it or the icon that shows when expanded.

BryceStevenWilley (approved):
LGTM, some small nits

docassemble/ALToolbox/misc.py
nonprofittechy: I considered that, but it was more confusing that way :) expand_icon is the one you click to expand it. It makes more sense to have the verbs match in my opinion.

docassemble/ALToolbox/misc.py
nonprofittechy: It's the default class. In the free group there is also `fab` for brands but that's not the most common use.

nonprofittechy: https://stackoverflow.com/questions/49913423/why-font-awesome-icon-class-is-different explains when fa was deprecated in fontawesome 5. Now there is fa-solid in fontawesome 6 which means the same thing but is clearer IMHO.

docassemble/ALToolbox/misc.py
BryceStevenWilley: >  expand_icon is the one you click to expand it

True, but the default here is `caret-down`, which is what's shown when it's already expanded before this PR. So you might be confused too ;) should probably fix that.

Even if I don't agree with the naming, if you added an explanation to the doc string I'd be fine.

nonprofittechy: Ah, I didn't mean to reverse it. I'll check on that.

nonprofittechy: I changed the adverbs to "closed" and "open" to match the CSS and your suggestion--also good catch on the accidental reversing of the default!

PR #118, opened 2022-09-22
More income module changes

### LINKED ISSUE #112: Income module: add an ALExpense class
It should model the code in https://github.com/nonprofittechy/docassemble-ssioverpaymentwaiver/blob/main/docassemble/ssioverpaymentwaiver/data/questions/financial_statement.yml

1. We can ask a question to say what category of expenses people have (via a checklist)
2. Then we ask people for each expense one at a time
3. We let people add duplicates of each expense too, so the computer can do the math

BryceStevenWilley:
* Made the ALPeriodicAmount class. Which is solely intended to house duplicate code between `ALExpense` and `ALIncome`. It straightens out the class hierarchy a bit, but it shouldn't be a concern for anyone but us, the module maintainers.
* adds the `move_checks_to_list`, to standardize the preferred way of gathering a big long list of potential expenses or incomes: by taking a checkboxes variable on a object and turning it into elements in a list with the `source` the same as the checkbox value
* made the ALJob class closer to the ALItemizedJob class. They shouldn't be different except for the Itemized part, so I made the employee an Individual
* Added the ALExpense and ALExpenses classes, as discussed previously (to avoid using the ALItemizedValueDict class)
* made slightly better names for the variables used for translation: `..._terms` and `..._terms_ordered`, instead of `list_dict`
* Moved over as many better default questions to `al_income.yml` from the Affidavit supplement when working on it.
  To be honest, there are still a lot of good default questions that we can't add, since you need to have access to the person containing the Job object. I suggest adding those questions to `AssemblyLine:ql_baseline.yml` for `users[0].jobs` and `ALIndividual.jobs` once this PR is in and we're content with the API
* went through all of `al_income_demo.yml` to fix some bugs and missing changes from this and #112.

Fixes #112

### DIFF
--- a/docassemble/ALToolbox/al_income.py
+++ b/docassemble/ALToolbox/al_income.py
@@ -27,7 +27,7 @@ __all__ = [
     "ALIncome",
     "ALIncomeList",
     "ALExpense",
-    "ALExpenses",
+    "ALExpenseList",
     "ALJob",
     "ALJobList",
     "ALAsset",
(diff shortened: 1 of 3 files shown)


nonprofittechy (changes requested):
Some minor nits, but one I think important for consistency is naming the class `ALExpenseList` instead of `ALExpenses`.

PR #127, opened 2022-11-03
Redefine the value of an item, not if it exists

BryceStevenWilley:
Asking only if an ItemizedValueDict entry exists won't ask the user the right questions; specifically letting them change the value of their dict. In this case, I've been setting up interviews to automatically define `exists` if asked for in this way, since usually I've been using `move_checks_to_list` to make objects already exist, yes / no. This means having `exists` and `value` re-edited doesn't ask two questions of the user.

Additionally, adds a default `revisit` question for the ALIncomeList object; otherwise edit buttons on review screens do nothing. 

Have tested with Affidavit Supplement, will likely merge without review, but feel free to give one post merge.


nonprofittechy (approved):
No concerns

Issue #123, opened 2022-11-02
Make the ThreePartsDate/BirthDate classes translatable


nonprofittechy:
Probably these are best translated with Docassemble's `word()` system

BryceStevenWilley:
Started this, but unfortunately, `word()` doesn't seem to be enough. It's likely because the labels are directly embedded in the `js_text` screen of the CustomDataType class. DA uses a metaclass on `CustomDataType` to [add all of the class attribute info to a register](https://github.com/jhpyle/docassemble/blob/b78a43e074d94d310951211aa5b84cce838a6457/docassemble_base/docassemble/base/functions.py#L5321), which only happens on server restart, so the context for the `word` function won't be right. This will likely take some additional features on docassemble's side, probably changing some things to reevaluate at runtime, or letting people make different classes for each language.

PR #128, opened 2022-11-03
Add labels to three part date

### LINKED ISSUE #15: Add proper aria to the BirthDate datatype
Currently with VoiceRSS, when you get to the birthdate field, it says "Input box" but there is no indication of which input of the three contains which input--month, day or year. It also only mentions one "input box".

Question: this could be correct if the aria is for the hidden input which accepts a date, properly typed. But might not work correctly with the jquery.

BryceStevenWilley:
Uses simple `label` attribute, and adds ids to each element. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label.

Also fixes other usability issues with the custom data type, including:
* better validation errors: right now, if you enter just one of the fields, you get an error like "10//" isn't a valid date. This isn't super clear of a message, so I added code that checks which of the fields are missing and directly shows them, e.g. "Enter a month and a year)
* If fields were partially entered, `NaN` would be placed in the fields. Now, only fields that can be parsed as numbers are re-placed in the fields.
* Started working on translation (#123), but couldn't get it to work fully, will need to continue elsewhere

From this I also opened #130 and #131, which I couldn't quite fix here without upstream fixes.

Fixes #15.


nonprofittechy:
Suggestion: can you combine with #123? I think just adding a `word()` wrapper gives us the scaffolding we need.

docassemble/ALToolbox/ThreePartsDate.py
```diff
@@ -112,24 +130,46 @@
...
+        return None
+    if all(empty_parts):
+        return word("Enter a month, a day, and a year")
+    else:
+        empty_words = [
+            empty_part[1]
+            for empty_part in zip(empty_parts, ["month", "day", "year"])
+            if empty_part[0]
+        ]
+    if len(empty_words) == 2:
```
nonprofittechy, line 154: This is probably fine but in gendered languages it's not simple enough to translate "a" the same for month, year and day. Might need separate sentences

docassemble/ALToolbox/ThreePartsDate.py
```diff
@@ -180,18 +223,21 @@ def validate(cls, item):
...
-                        word("Answer with a <strong>date of birth</strong>")
+                        word(
+                            f"Answer with a <strong>date of birth</strong> ({date} is in the future)"
+                        )
                     )
             else:
-                msg = (
-                    f"{ item } {word('is not a valid <strong>date of birth</strong>')}"
+                msg = check_empty_parts(
+                    item, default_msg="is not a valid <strong>date of birth</strong>"
```
nonprofittechy, line 240: Again this might not be a translatable fragment. Just worth noting before we start adding any entries in word.yml

nonprofittechy (commented):
A few notes on translation in the body

docassemble/ALToolbox/ThreePartsDate.py
BryceStevenWilley: This on is a bit more complicated though, as it's directly concatenated with the invalid date. We can't know exactly where the date is going to be.

Is it normal for the result of the `word` function to still have formatting variables, like `"The number is  is {my_var}"`? That would be doable I guess.

PR #144, opened 2023-01-24
Remove `net` attribute of ALJob

BryceStevenWilley:
Replace it with the `deduction` attribute. This leaves `ALJob` with a `value` and `deduction` attributes, which can be used to calculate the `net_total()` instead of directly asking people for their net income, something they might know but would have to look up and do math to confirm exactly.


BryceStevenWilley:
Going to merge. Likely won't update Pypi with a release until #145 is merged, but working on finishing that today.

PR #145, opened 2023-01-25
Tutorial changes

BryceStevenWilley:
Goes with the changes I'm making in https://github.com/SuffolkLITLab/docassemble-AssemblyLine-documentation/pulls/310, and just adds a lot of polish that I noticed was lacking when working on the demo and the documentation (since much of the development was done with overriding questions in the ALAffidavit interview, I didn't have a good idea of what the base experience looked like).

If you want an idea of what's changed, it should be summarized in the commit messages, but it's a bit scattershot. Willing to answer any questions about changes.


docassemble/ALToolbox/data/questions/al_income.yml
```diff
@@ -117,6 +117,26 @@ subquestion: |
 
   ${ x.add_action() }
 ---
+generic object: ALItemizedJob
+continue button field: x.to_add.revisit
+question: |
+  Edit incomes
+subquestion: |
+  ${ x.to_add.table }
```
plocket, line 125: Seeing a few things here I haven't caught before or that I'm perhaps misunderstanding. Sorry either way.

There are no review screens in the demo from what I can tell, so people don't get to see the review screens for the items. I didn't quite have time to write one myself, sorry, so I didn't get to see the new tables. I can try to do that another time. My comments here are based just on reading code.

1. Doesn't break this PR, but is tangentially relevant to it: I'm not sure there's any `x.table` for the type of object that this is, which is an `ALItemizedValueDict`.
2. Future issue: I see these edit screens for ALItemizedJobList as well, but I don't think that class has those properties, just the jobs inside have those properties.
3. Future issue: I think I'm actually seeing some missing tables in the code in general and at least one repeated item - the table for `ALExpenseList`, which I see at lines 614 and 161.

plocket (changes requested):
Some really great catches and improvements in here!

I did find at least one thing that needs changing (syntax), and other things that could be improvements to this PR. There are also some questions and some things perhaps for future PRs.

Here are a few that I didn't find a good place to put in the code comments:
1. On the final screen of the demo, `itemized_job.output` seems to be requiring the `.name` attribute, which causes an error. [Second to last line in the block](https://github.com/SuffolkLITLab/docassemble-ALToolbox/blob/98a1eabdc9d3753c8efffd28e560fb64ebefdf73/docassemble/ALToolbox/data/questions/al_income_demo.yml#L410).
2. An itemized job output:
```
Job’s .to_add as a string: [ [ “part time”, “100.00” ] ]
```
Is that the intended string output?

docassemble/ALToolbox/al_income.py
```diff
@@ -545,8 +553,8 @@ class ALExpenseList(ALIncomeList):
     A list of expenses
 
     * each element has a:
+        * value
         * source
-        * owner
```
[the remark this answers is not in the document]
BryceStevenWilley: Expenses don't set `.owner` in any questions, and it generally doesn't make sense for an expense to have an "owner". Might be the case on specific lines of credit or loans, but not general expenses like groceries.

docassemble/ALToolbox/data/questions/al_income_demo.yml
```diff
@@ -36,113 +39,106 @@ id: interview order
...
     al_value_list.gather()
   
-  before_end_so_answers_are_saved
   end
 ---
+question: ${ i }
+subquestion: |
+  The next screens will demonstrate how a ${ i } will appear in an interview.
+
+  % if i == "ALIncome"
```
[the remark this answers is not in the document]
BryceStevenWilley: Thanks, I just found that this morning.

docassemble/ALToolbox/al_income.py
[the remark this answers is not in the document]
plocket: Yes, exactly. I'm saying ALIncomeList may have the same erroneous documentation.

docassemble/ALToolbox/data/questions/al_income.yml
BryceStevenWilley: > Couldn't even trigger it with itemized_job.to_add.revisit

ALItemizedJob doesn't have a revisit screen itself. It splits up better as a row in the ALItemizedJobList revisit screen; then if you edit the job, you can change the job title, employer info, and then get two separate screens for editing incomes from the ValueDict and editing deductions from the ValueDict. TBH most people won't be using ALItemizedJob by itself: they should be using the list, which does have a revisit screen.

> I'm not sure there's any x.table for the type of object that this is, which is an ALItemizedValueDict

There is, right above this, lines 111-118.

> I see these edit screens for ALItemizedJobList as well, but I don't think that class has those attributes, just the jobs inside have those attributes

I assume you're talking about the `ALItemizedJobList.table` (lines 468- 485)? That's how tables work: everything in the edit list are attributes of the things in the list.

> I think I'm actually seeing some missing tables in the code in general, and at least one repeated item - the table for ALExpenseList, which I see at lines 614 and 161.

Thanks! The repeated item is addressed now, and I added tables for ALIncomeList and ALAssetList. I don't have any examples of ALVehicleList or ALSimpleValueList yet, so I won't add those until I have a solid understanding of them (likely when I add those docs).

plocket: > lines 468- 485

If those are the relevant tables, I think line 480 might need adjustment. It uses `x.table` instead of `x.to_add.table`:
```
generic object: ALItemizedJobList
continue button field: x.to_add.revisit
question: |
  Edit incomes from ${ x.source }
subquestion: |
  ${ x.table }
  ${ x.add_action() }
```
From what I can see, this is different than the `x.to_subtract.revisit` block.

docassemble/ALToolbox/al_income.py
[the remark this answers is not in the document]
BryceStevenWilley: Ah, no, that's intentional. The full text there is "It can make use of these attributes and methods in its items", meaning owner is an attribute of ALIncome.

docassemble/ALToolbox/data/questions/al_income.yml
BryceStevenWilley: Another good catch. These were definitely bugged and not appearing, which is why I added these blocks (120 - 139). I'm going to delete them, and maybe reorganize things a bit.

BryceStevenWilley:
> Job’s .to_add as a string: [ [ “part time”, “100.00” ] ]

I believe so. From the docs of that `__str__` method,

> Returns a string of the dictionary's key/value pairs in a list. E.g.
        "['federal_taxes': '2500.00', 'wages': '15.50']"
        
But then adds tuples of the key values to the list to turn into the string. I'm not sure the ultimate intention, but a list of lists is understandable IMO. I'll change the docs, but let me know if you have something else in mind.

BryceStevenWilley:
Should be ready for a re-review. The only thing that I didn't do was add review screens to the demo. I'll consider it later, but I've spent a lot of time on the docs so far, and that would make things take longer.

plocket:
`all of the above` is currently giving me an "overwrites previous label" error. I've put a question out there as I couldn't track it down so far.

docassemble/ALToolbox/data/questions/al_income_demo.yml
```diff
@@ -31,118 +34,118 @@ objects:
...
-  * `ALJob` covers all the same functionality as `ALIncome`.
-  * `ALJobList` covers all the same functionality as `ALJob` and `ALIncome`.
-  * `ALAsset` covers all the same functionality as `ALIncome`.
-  * `ALAssetList` covers all the same functionality as `ALAsset` and `ALIncome`.
-  * `ALVehicleList` covers all the same functionality as `ALVehicle`.
-  * `ALSimpleValueList` covers all the same functionality as `ALSimpleValue`.
-  
-  Each item will only demo its own capabilities and those of the classes from which it inherits. It does not show additional features that you can add. For example, an individual ALAsset won't ask for a market value as it has no methods for calculating a market value. An ALAssetList, though, will demo summing up market values using the filters it has in its method.
-
+  This interview let's you try out functionalities from the `al_income` module. You can also
```
plocket, line 113:
```suggestion
  This interview let's you try out functionalities from the `al_income` module. You can also
```
```suggestion
  This interview lets you try out functionalities from the `al_income` module. You can also
```

docassemble/ALToolbox/data/questions/al_income_demo.yml
```diff
@@ -403,11 +410,11 @@ content: |
   
   Name | Calculation type | Frequency | Gross amount
   - | - | - | -
-  % for name, item in x.to_add.complete_elements().items():
-  ${ job_items_names.get(name, name) } | Gross | Monthly | ${ currency(x.gross_total( source=name, times_per_year=12 )) }
+  % for source, item in x.to_add.complete_elements().items():
+  ${ job_items_names.get(source, source) } | Gross | Monthly | ${ currency(x.gross_total( source=source, times_per_year=12 )) }
   % endfor
   % for name, item in x.to_subtract.complete_elements().items():
```
plocket, line 416: I think this last for loop is using `source` again instead of `name`. When I went through the interview, I had two items to subtract.
```suggestion
  % for source, item in x.to_subtract.complete_elements().items():
```

I had two deductions and I got two duplicate rows, both showing `to_add` at the final screen:

dup_rows

plocket (changes requested):
Sorry this is dragging out. I think some of these things are not/have not been from the changes you've made here. Kudos for taking this on.

There were a couple errors. One that I remarked on in its line. The other is for filling out an income list - I get an error when I select "Other". `There was a reference to a variable 'al_income_list[0].source' that could not be looked up`. Maybe it's worth leaving that option off for now?

Sorry to request changes again.

nonprofittechy:
> `all of the above` is currently giving me an "overwrites previous label" error. I've put a question out there as I couldn't track it down so far.
> 
> [Answer: Our da package is old and we need to update.] [Done and it works now.]

Let's avoid adding features that require the newest DA release

BryceStevenWilley:
> Let's avoid adding features that require the newest DA release

Removing it then, since our "manual" all of the above wasn't working with url_args.

plocket (approved):
Looking good from what I can see! The caveat is that I think we'll leave some things to test more throughly when they're relevant, like review screen things, because it's hard to test them right now.

Congrats! That was a ton to wade through.

PR #151, opened 2023-01-30
ThreePartDate: put str in error message, not date

BryceStevenWilley:
We've seen some errors in DA, where formatting a date into a string tries to get info from the current thread (the interview title?!) that it doesn't have, and will error. This should fix that issue, but just showing the date ppl gave to us.

Keeping this draft until I can:
* recreate the issue on apps-dev
* confirm that this branch fixes the issue


BryceStevenWilley:
Have confirmed it still exists, even if I can't recreate. Going to merge, and see if this fixes for others.

PR #168, opened 2023-03-18
Add demo for multiselect (still not too modular yet)

nonprofittechy:


nonprofittechy:
Going to merge just so that there's a simple example of the multiselect code working in its current state. The example in TCLP is too complicated to copy and modify

PR #193, opened 2023-09-20
Enforce people entering numbers for hours

BryceStevenWilley:
We had `input type: number`, which doesn't exist, and so was doing nothing. Changed to `datatype: number`, and added validation messages that should show if the user enters something that's not a number.

### DIFF
--- a/docassemble/ALToolbox/al_income.py
+++ b/docassemble/ALToolbox/al_income.py
@@ -1028,6 +1028,14 @@ class ALItemizedJob(DAObject):
         else:
             frequency_to_use = self.times_per_year
 
+        # NOTE: fixes a bug that was present < 0.8.2
+        try :
+          hours_per_period = Decimal(self.hours_per_period)
+        except:
+          log(word("Your hours per period need to be just a single number, without words"), "danger")
+          delattr(self, "hours_per_period")
+          self.hours_per_period # Will cause another exception
+
         # Both the job and the item itself need to be hourly to be
         # calculated as hourly
         is_hourly = self.is_hourly and hasattr(item, "is_hourly") and item.is_hourly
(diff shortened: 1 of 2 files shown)


BryceStevenWilley:
A bit more context here: there are production errors that have the following error:

```
  File "/usr/share/docassemble/local3.10/lib/python3.10/site-packages/docassemble/ALToolbox/al_income.py", line 1246, in gross_total
    total += job.gross_total(
  File "/usr/share/docassemble/local3.10/lib/python3.10/site-packages/docassemble/ALToolbox/al_income.py", line 1082, in gross_total
    total += self._item_value_per_times_per_year(
  File "/usr/share/docassemble/local3.10/lib/python3.10/site-packages/docassemble/ALToolbox/al_income.py", line 1039, in _item_value_per_times_per_year
    value * Decimal(self.hours_per_period) * Decimal(frequency_to_use)
decimal.InvalidOperation: []
```

I hadn't been able to fix this without looking at the inputs that people give, because the error doesn't have enough information. But I discovered it was because people were entering full sentences in that field, which obviously causes a conversion error.

I looked at trying to patch `al_income.py` specifically, but I do think that throwing an exception is the right thing to do in that case; the class just shouldn't have full text strings for those values, and if it does at that point, I don't think there's anything we could do [^1].


[^1]: We could try to delete the offending attribute and call `log(..., "error")` which shows text to users directly, but if we did, the user would just get sent back to the screen where they entered the information, and some text at the top. We don't know what the field name on that screen is, or anything else. If we're interested in that fix, let me know and I could make it.

BryceStevenWilley:
@nonprofittechy , @plocket Realized that I forgot to mention this fix I made last week on deep-dive today.

The thing I wanted to float was if we thought that the solution [I mentioned here](https://github.com/SuffolkLITLab/docassemble-ALToolbox/pull/193#user-content-fn-1-17126e4368f845d88a08204ecfa79627) is a good idea; otherwise, existing interviews where people have entered non-numbers will continue to be broken.

nonprofittechy (approved):

BryceStevenWilley:
Tested and it works fine. TBH I have more concerns about the number validation UX messages than I do about sending people back.

> I suspect that the existing broken interviews are just abandoned forever though, so it's not necessarily urgent.

That's true, it's not urgent, but it is a nice safety feature to have as well IMO, in case other override the question.

PR #198, opened 2023-10-11
Seasonal income classes

plocket:
It's messy, as we knew it would be. I'd appreciate feedback if folks have the time. I've created an [experimental yml](https://github.com/plocket/docassemble-ALSeasonalIncome/blob/main/docassemble/ALSeasonalIncome/data/questions/seasonal_temp.yml) for implementation, but I'm unable to get it working in the ways I'd expect it to work. I keep triggering default questions instead of month questions. Missing something fundamental.

I'm concerned about `.is_hourly` and `.hours_per_period`. Somewhere along the way they started getting required. I tried to restore them to their previous optional status as they were giving me trouble for a non-hourly job, but I may be missing context. May need live discussion.

It also looks like the `holidays` dependency has something going on as tests are failing on it.

### DIFF
--- a/docassemble/ALToolbox/al_income.py
+++ b/docassemble/ALToolbox/al_income.py
@@ -941,6 +941,7 @@ class ALItemizedJob(DAObject):
     - Overtime at a second hourly rate
     - Tips earned during that time period
     - A fixed salary earned for that pay period
+    - Income and deductions from a seasonal job
     - Union Dues
     - Insurance
     - Taxes
@@ -960,6 +961,8 @@ class ALItemizedJob(DAObject):
         user earns on an hourly basis, rather than for the full time period
     .hours_per_period {int} (Optional) If the job is hourly, how many hours the
         user works per period.
+    .is_seasonal {bool} (Optional) Whether the job's income changes drastically
+        during different times of year.
     .employer {Individual} (Optional) Individual assumed to have a name and,
         optionally, an address and phone.
     .source {str} (Optional) The category of this item, like "public service".
@@ -985,8 +988,6 @@ class ALItemizedJob(DAObject):
 
     def init(self, *pargs, **kwargs):
         super().init(*pargs, **kwargs)
-        # if not hasattr(self, "source") or self.source is None:
-        #    self.source = "job"
         if not hasattr(self, "employer"):
             if hasattr(self, "employer_type"):
                 self.initializeAttribute("employer", self.employer_type)
@@ -1043,7 +1044,7 @@ class ALItemizedJob(DAObject):
 
         # Both the job and the item itself need to be hourly to be
         # calculated as hourly
-        is_hourly = self.is_hourly and hasattr(item, "is_hourly") and item.is_hourly
+        is_hourly = hasattr(self, "is_hourly") and self.is_hourly and hasattr(item, "is_hourly") and item.is_hourly
         value = item.total()
 
         # Use the appropriate calculation
--- a/docassemble/ALToolbox/al_income.py
+++ b/docassemble/ALToolbox/al_income.py
@@ -36,8 +36,8 @@ __all__ = [
     "ALVehicleList",
     "ALSimpleValue",
     "ALSimpleValueList",
-    "ALItemizedValue",
-    "ALItemizedValueDict",
+    "ALItemizedValue",  # Rationale for exporting this?
+    "ALItemizedValueDict",  # Rationale for exporting this?
     "ALItemizedJob",
     "ALItemizedJobList",
 ]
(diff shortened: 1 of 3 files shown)


docassemble/ALToolbox/al_income.py
```diff
@@ -958,12 +959,18 @@ class ALItemizedJob(DAObject):
...
-        user works per period.
+    .hours_per_period {float | Decimal} (Optional) If the job is hourly, how
+        many hours the user works per period.
+    .is_seasonal {bool} (Optional) Whether the job's income changes drastically
+        during different times of year.
     .employer {Individual} (Optional) Individual assumed to have a name and,
         optionally, an address and phone.
     .source {str} (Optional) The category of this item, like "public service".
-        Defaults to "job".
+    .months {ALItemizedJobMonthList} Automatically exist, but they won't be used
```
nonprofittechy, line 969: Question: is this meant to be `months`  or `paystubs`?

docassemble/ALToolbox/al_income.py
```diff
@@ -1028,26 +1038,28 @@ def _item_value_per_times_per_year(
...
-        except:
-            log(
-                word(
-                    "Your hours per period need to be just a single number, without words"
-                ),
-                "danger",
-            )
-            delattr(self, "hours_per_period")
-            self.hours_per_period  # Will cause another exception
-
```
nonprofittechy, line 1043: I don't think this is ready to remove yet. @BryceStevenWilley ?

docassemble/ALToolbox/data/questions/al_income_demo.yml
```diff
@@ -4,6 +4,66 @@ metadata:
...
+      .source {str} (Optional) The category of this item, like "public service".
+      """
+      def init(self, *pargs, **kwargs):
+          super().init(*pargs, **kwargs)
+          self.ask_number = True
+          self.target_number = 12
+          month_names = [
+              "january", "february", "march", "april", "may", "june",
+              "july", "august", "september", "october", "november"
+          ]
```
nonprofittechy, line 29: Same question as above. I'm just curious if this needs to be limited to chunking by month. Biweekly pay is a bit more common for most folks.

nonprofittechy (commented):
This is looking pretty reasonable to me. I'm just wondering if months is the right level to chunk on. It's still going to require most folks to do their own math, isn't it? Since people will have bi-weekly paystubs most of the time, not monthly.

docassemble/ALToolbox/al_income.py
```diff
@@ -1317,3 +1333,37 @@ def net_total(
...
+        kwargs['source'] = kwargs.get('source', "months")
+        kwargs['object_type'] = kwargs.get('object_type', ALItemizedJobMonth)
+        kwargs['ask_number'] = kwargs.get('ask_number', True)
+        kwargs['target_number'] = kwargs.get('target_number', 12)
+        
+        kwargs['add_months'] = kwargs.get('add_months', False)
+        super().init(*pargs, **kwargs)
+        
+        month_names = [
+            "january", "february", "march", "april", "may", "june",
```
nonprofittechy, line 1364: Assuming we really do need a month, should this just be an integer to simplify translation? Can use `format_date` to display it as text in the interface. I'm not usually seeing people store months as a list of hardcoded strings.

docassemble/ALToolbox/data/questions/al_income_demo.yml
plocket: That's a good question. @CaroRob and I have been thinking about that discussion. I had hesitated to go beyond months since some seemed reluctant about getting even this granular, but if folks are open to that discussion, I am too.

docassemble/ALToolbox/al_income.py
plocket: I think it depends on where we go with this and how granular we let people get (as you discuss elsewhere) and how much we want to let them customize.

docassemble/ALToolbox/data/questions/al_income_demo.yml
nonprofittechy: Think about whether limiting it to months helps achieve any of the goals. It might, but I'm not sure. I think you discussed letting people avoid math. Monthly chunking doesn't seem like it would save any math.

Maybe letting users enter a date range? E.g., teachers usually have 9 month contracts, although they don't always have the choice to get paid over 9 months rather than 12.

plocket: I've made it more flexible. Will wait to fully adjust the demo till I get the go-ahead. Here's a sample file that shows the most basic application (you must answer that, yes, your job is seasonal). It's .txt and you have to have to change it to .yml because GitHub won't let me upload a .yml.

[seasonal_income_demo.txt](https://github.com/SuffolkLITLab/docassemble-ALToolbox/files/13462189/seasonal_income_demo.txt)

PR #205, opened 2023-11-09
Add to the demo to show using a placeholder for the phone numbers

### LINKED ISSUE #201: datatype: al_international_phone phone number fields confused some testers
In user testing, a couple people have gotten confused by the placeholder text in these **datatype: al_international_phone** phone number fields. They thought they were already filled in. I wonder if having (000) 000-0000 would be better instead of what looks like a real phone number?

plocket:
Informs #201


nonprofittechy:
This is fine but I don't think it fixes the original issue. Probably the default when no "hint" is provided should be to have no placeholder.

I can't actually think of a time when we want to use placeholder text. It's actively discouraged in our style guide.

PR #212, opened 2023-11-24
Remove unnecessary quote

BryceStevenWilley:
Ran into this when using the Dashboard review generator, and there was an extra quote at the end of the last block. Isn't necessary, and makes the resulting DA YAML have a subtle bug.

Will test on dev, and remove if it fixes the issue.


plocket (approved):
Not sure you need a review, but it looks good to me!

PR #216, opened 2023-12-04
Make collapse/display templates work in dark mode

BryceStevenWilley:
In dark mode, the display templates still have a light background, causing contrast issues with the text inside, and in general stands out far too much against the page background. This adds CSS to change the `bg-light` class for the collapse and display templates to the dark-mode colors when `data-bs-theme=dark`.

See https://github.com/jhpyle/docassemble/pull/718 for a discussion on a suggested fix for a similar issue, which resulted in the suggested workaround here.

An image of the dark mode display template without this PR:

Screenshot from 2023-12-04 18-46-26

An image of the dark mode display template with this PR:

Screenshot from 2023-12-04 18-45-29

### DIFF
--- a/docassemble/ALToolbox/data/static/collapse_template.css
+++ b/docassemble/ALToolbox/data/static/collapse_template.css
@@ -29,3 +29,12 @@
 .al_display_template a.collapsed .pdcaretclosed {
     display: inline;
 }
+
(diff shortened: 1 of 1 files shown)


nonprofittechy (approved):
Great! Lgtm

Issue #164, opened 2023-02-22
Make dates accessible


plocket:
We do have labels for each of our own inputs, but I just realized that we don't have a matching `name` for the whole field label's `for`. I'm not sure exactly how to juggle that - can we move the label that's on the original date field, or does that break submitting the data? Should we change the field label's `for` to something we create ourselves? Do we need a `legend` to contain stuff to help the label be associated with those multiple fields? Maybe look at how checkboxes are made accessible.

BryceStevenWilley:
Related to #214

plocket:
By that, I'm guessing we don't need to associate the original date field with that group of fields, right?

BryceStevenWilley:
I don't think so, they just need to be clearly labeled (i.e. all of the fields should still be related to the original field label somewhere, but we don't care about the original field itself)

plocket:
Sorry I'm not being clear. That's the exact question I'm asking. I'm not sure we're currently associating each field with the original field label. We're associating them with their own labels - "month", "day", and "year", but I'm not sure that we're associating them with their main label, e.g. "The date you bought the house"

BryceStevenWilley:
Ah, I see. I do think we want that; otherwise if there are multiple date inputs on a page, there will be multiple fields with the same labels ("Month", etc.), which can cause issues with accessibility tools that list all of the fields on a page and such.

I talk about bit more about how this could be done in https://github.com/SuffolkLITLab/docassemble-ALToolbox/pull/230#pullrequestreview-1780837633

plocket:
The most specific section from those comments (since I think this is possibly a harder problem to solve and would rather keep the discussion in here):

> I think that we could make the whole widget a [fieldset](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset), with that label becoming the `legend`. I manually edited that HTML in the browser tools, and it's more clearly associated with the inputs, which is great, essentially just changing the `div.da-field-container-datatype-ThreePartsDate` to a `fieldset`, and changing the `label` tag to a `legend`. My HTML looks like this:
> 
> ```html
> <fieldset class="da-container da-form-group row da-field-container da-field-container-datatype-ThreePartsDate">
>     <legend for="ZGF0ZV9vZl9ncmFkdWF0aW9u" class="col-md-4 col-form-label da-form-label datext-right">
>         When will your high school graduation be?
>     </legend>
>     <div class="col-md-8 dafieldpart">
[...]

PR #230, opened 2023-12-12
Improve accessibility by associating `for` with `id` instead of `name`

### LINKED ISSUE #214: for/id associations for ThreePartDate
From VT's accessibility audit, relating to the ThreePartDateInput.

> On step 11, the “Month”, “Day”, and “Year” < label > elements are disassociated from their inputs
(seemingly deliberately, judging by the word “ignore” in their for attributes). As a result, screen reader
users cannot hear these labels. We recommend creating proper for / id associations. 3-Moderate

Here's an example chunk of the ThreePartDate HTML for reference.

```html
<div class="col-md-8 dafieldpart">
    <div class="al_three_parts_date form-row row">
        <div class="col col-month">
            <label for="_ignore_ZGF0ZV9vZl9ncmFkdWF0aW9u_month">Month</label>
            <select class="form-select al_field month ZGF0ZV9vZl9ncmFkdWF0aW9u" id="ZGF0ZV9vZl9ncmFkdWF0aW9u_month" name="_ignore_ZGF0ZV9vZl9ncmFkdWF0aW9u_month" aria-describedby="ZGF0ZV9vZl9ncmFkdWF0aW9u-error" aria-invalid="false">
                <option value=""></option>
                <option value="01">January</option>
                <option value="02">February</option>
                ...
            </select>
        </div>
        <div class="col col-3 col-day">
            <label for="_ignore_ZGF0ZV9vZl9ncmFkdWF0aW9u_day">Day</label>
            <input class="form-control al_field day ZGF0ZV9vZl9ncmFkdWF0aW9u" type="number" min="1" inputmode="numeric" id="ZGF0ZV9vZl9ncmFkdWF0aW9u_day" name="_ignore_ZGF0ZV9vZl9ncmFkdWF0aW9u_day" aria-describedby="ZGF0ZV9vZl9ncmFkdWF0aW9u-error">
        </div>
        <div class="col col-3 col-year">
            <label for="_ignore_ZGF0ZV9vZl9ncmFkdWF0aW9u_year">Year</label>
            <input class="form-control al_field year ZGF0ZV9vZl9ncmFkdWF0aW9u" type="number" min="1" inputmode="numeric" id="ZGF0ZV9vZl9ncmFkdWF0aW9u_year" name="_ignore_ZGF0ZV9vZl9ncmFkdWF0aW9u_year" aria-describedby="ZGF0ZV9vZl9ncmFkdWF0aW9u-error">
        </div>
        <div id="al_ZGF0ZV9vZl9ncmFkdWF0aW9u_error" class="al_error"></div>
    </div>
    <input value="02/3/2040" alt="Input box" class="form-control daThreePartsDate" data-almin="11/28/2023" data-almax="12/31/2040" type="hidden" name="ZGF0ZV9vZl9ncmFkdWF0aW9u" id="ZGF0ZV9vZl9ncmFkdWF0aW9u" style="display: none;" aria-hidden="true">
</div>
```

I'll have to look into what DA and HTML normally do with `name` compared to `id`, but I might need your help / perspective when building this @plocket

plocket:
Closes #214. No rush on my part. I've tested it a bit myself to make sure basic behavior hasn't changed, though I don't think I have the tools I need to test the accessibility properly.

Not sure if we want to increment those package versions.

### DIFF
--- a/docassemble/ALToolbox/ThreePartsDate.py
+++ b/docassemble/ALToolbox/ThreePartsDate.py
@@ -175,7 +175,7 @@ function create_month(date_id) {{
   */
   var $col = $('<div class="col col-month">');
   
-  var id = '_ignore_' + date_id + '_month';
+  let id = '_ignore_' + date_id + '_month';
   // '_ignore' in the name prevents the field from being submitted, avoiding a da error
   let name =  id;
   
--- a/docassemble/ALToolbox/ThreePartsDate.py
+++ b/docassemble/ALToolbox/ThreePartsDate.py
@@ -179,7 +179,7 @@ function create_month(date_id) {{
   // '_ignore' in the name prevents the field from being submitted, avoiding a da error
   let name =  id;
   
-  var $label = $('<label for="' + id + '">{month}</label>');
+  let $label = $('<label for="' + id + '">{month}</label>');
   $col.append($label);
   
   // aria-describedby is ok to have, even when the date-part error is
--- a/docassemble/ALToolbox/ThreePartsDate.py
+++ b/docassemble/ALToolbox/ThreePartsDate.py
@@ -130,16 +130,16 @@ function create_date_part({{type, date_id}}) {{
   * @returns undefined
   */
   var $col = $('<div class="col col-3 col-' + type + '">');
-  var id = date_id + '_' + type;
-  // '_ignore' prevents the field from being submitted, avoiding a da error
-  let name =  '_ignore_' + id;
+  var id = '_ignore_' + date_id + '_' + type;
+  // '_ignore' in the name prevents the field from being submitted, avoiding a da error
+  let name = id;
   
   // For python formatting, need to have {{day}} and {{year}}
   let $label = '';
   if (type === 'day') {{
-    $label = $('<label for="' + name + '">{day}</label>');
+    $label = $('<label for="' + id + '">{day}</label>');
   }} else {{
-    $label = $('<label for="' + name + '">{year}</label>');
+    $label = $('<label for="' + id + '">{year}</label>');
   }}
   $col.append($label);
   
@@ -175,11 +175,11 @@ function create_month(date_id) {{
   */
   var $col = $('<div class="col col-month">');
   
-  let id = date_id + '_month';
-  // '_ignore' prevents the field from being submitted, avoiding a da error
-  let name =  '_ignore_' + id;
+  var id = '_ignore_' + date_id + '_month';
+  // '_ignore' in the name prevents the field from being submitted, avoiding a da error
+  let name =  id;
   
-  let $label = $('<label for="' + name + '">{month}</label>');
+  var $label = $('<label for="' + id + '">{month}</label>');
   $col.append($label);
   
   // aria-describedby is ok to have, even when the date-part error is
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ setup(name='docassemble.ALToolbox',
       url='https://suffolklitlab.org/docassemble-AssemblyLine-documentation/docs/framework/altoolbox',
       packages=find_packages(),
       namespace_packages=['docassemble'],
-      install_requires=['holidays>=0.25', 'pandas>=2.0.3'],
+      install_requires=['holidays>=0.27.1', 'pandas>=2.0.3'],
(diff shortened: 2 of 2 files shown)


nonprofittechy (approved):
Just need to fix formatting; I'll merge that PR

plocket:
I think Bryce wanted to test some things before merging this

nonprofittechy:
> I think Bryce wanted to test some things before merging this

I meant merge the PR from actions/Black, but that doesn't exist. Looks like we need to point to Bryce's new action in ALActions for formatting; the one this repo points to is broken.

docassemble/ALToolbox/ThreePartsDate.py
```diff
@@ -175,11 +175,11 @@
   */
   var $col = $('<div class="col col-month">');
   
-  let id = date_id + '_month';
-  // '_ignore' prevents the field from being submitted, avoiding a da error
-  let name =  '_ignore_' + id;
+  var id = '_ignore_' + date_id + '_month';
```
BryceStevenWilley, line 178: Any reason you changed `let` to `var` here, but not elsewhere? IMO it's either we do need it for compatibility, or we don't.

BryceStevenWilley (changes requested):
Got to testing it out, and it does seem to associate the "Month" / "Day" / "Year" labels better with their inputs, so success! However, I think the conversation we started having in #164 comes into play too, as there are some issues with the original field label.

In Chromium, I get this error:

```
The label's for attribute doesn't match any element id. This might prevent the
browser from correctly autofilling the form and accessibility tools from working correctly.

To fix this issue, make sure the label's for attribute references the correct id of a form field.
```

which happens because we set that input to be hidden, so most accessibility tools won't include it as a part of the page's content, and will assume that the label points to nothing.

I think that we could make the whole widget a [fieldset](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset), with that label becoming the `legend`. I manually edited that HTML in the browser tools, and it's more clearly associated with the inputs, which is great, essentially just changing the `div.da-field-container-datatype-ThreePartsDate` to a `fieldset`, and changing the `label` tag to a `legend`. My HTML looks like this:

```html
<fieldset class="da-container da-form-group row da-field-container da-field-container-datatype-ThreePartsDate">
    <legend for="ZGF0ZV9vZl9ncmFkdWF0aW9u" class="col-md-4 col-form-label da-form-label datext-right">
        When will your high school graduation be?
    </legend>
    <div class="col-md-8 dafieldpart">
        (unchanged)
        ...
</fieldset>
```

I'm not sure how much harder these changes would be; if you think they're a lot harder, we can merge this and take another shot at them later (and to be extra clear, these issues are also present in main, so there's nothing wrong with this PR). But if you think that it'd be easy to add while this is fresh in our minds, I think it'd be a significant improvement.

docassemble/ALToolbox/ThreePartsDate.py
plocket: You're right, we should be able to use `let` now. It's just an old habit. I'll change it back 👍

plocket:
> I'm not sure how much harder these [fieldset] changes would be; if you think they're a lot harder, we can merge this and take another shot at them later

I do believe the `fieldset` conversation is a more complex one and [added to the discussion in #164](https://github.com/SuffolkLITLab/docassemble-ALToolbox/issues/164#issuecomment-1856049920). We should probably save it for a separate discussion and PR. Maybe a deep dive? I've added it to the list in the doc for whenever the next deep dive happens.

docassemble/ALToolbox/ThreePartsDate.py
```diff
@@ -175,11 +175,11 @@
...
   
-  let id = date_id + '_month';
-  // '_ignore' prevents the field from being submitted, avoiding a da error
-  let name =  '_ignore_' + id;
+  let id = '_ignore_' + date_id + '_month';
+  // '_ignore' in the name prevents the field from being submitted, avoiding a da error
+  let name =  id;
   
-  let $label = $('<label for="' + name + '">{month}</label>');
+  var $label = $('<label for="' + id + '">{month}</label>');
```
BryceStevenWilley, line 182:
```suggestion
  let $label = $('<label for="' + id + '">{month}</label>');
```

PR #232, opened 2023-12-14
Make display_template look like a daterm

### LINKED ISSUE #208: Change `display_template()` links to green words color with same dotted underline
We saw on usability testing for MOTenantHelp that `display_template()` currently shows as a blue underlined hyperlink (as if a URL) with the only difference being that it has a caret. One user hesitated to click on the template because they thought it would open a new tab and would cause their wifi to be slow.

See https://piccalil.li/tutorial/a-progressive-disclosure-component/

See https://github.com/LemmaLegalConsulting/docassemble-MOHUDEvictionProject/issues/457

BryceStevenWilley:
They serve the same function (tap to get more information, and they don't navigate you away from the page), so it makes sense to brand them the same.

I'm not quite used to it yet, but I do think I will with a bit of time. Also have to keep in mind that most users aren't coming back again and again, they use our side a few times and don't return.

Fixes #208

Screenshot from 2023-12-13 20-46-57

### DIFF
--- a/docassemble/ALToolbox/data/static/collapse_template.css
+++ b/docassemble/ALToolbox/data/static/collapse_template.css
@@ -5,6 +5,13 @@
   padding-bottom: 1rem;
 }
 
+.al_collapse_template a.al_toggle,
+.al_display_template a.al_toggle {
+  color: #3b842c; /* Same as daterm in Assembly Line */
+  text-decoration: dashed;
+  text-decoration-line: underline;
+}
+
 .al_collapse_template .collapse,
 .al_display_template .collapse {
     margin-bottom: 1rem;


nonprofittechy (approved):
You're right it will take a bit to get used to, but I think it's the right change. Looks good to merge

miabonardi:
Is this only supposed to work on collapse_template? I am seeing difference between collapse and display templates

Screen Shot 2023-12-14 at 1 06 32 PM

ekressmiller:
For us this comes out looking beautiful in Firefox:
image

But kind of hard to read in Chrome:
image

I think the chunky dashes underneath become more of an issue with longer stretches of text like in the template headers.  I'll think about it/look into what we can do on our servers but just wanted to mention it!

BryceStevenWilley:
Good spot, I agree that the Chrome one isn't very readable. You can use some CSS in your interview to adjust it:

```css
.al_collapse_template a.al_toggle, .al_display_template a.al_toggle {
  text-underline-offset: 3px;
}
```

Here's a screenshot: Left is chrome, right is firefox.

Screenshot from 2024-01-16 23-07-19

There's also `text-decoration-thickness`. Not sure how widely supported both are, but it might be good to space out the Chrome one a bit in AL core. Annoying that the browser's have horrible readability defaults.

(I didn't check, but you might want to see what the terms look like in chrome. I'm pretty sure I used the same CSS as the terms used, and if you fix this CSS, you'll want to fix the term CSS as well).

ekressmiller:
Thanks Bryce!

Issue #191, opened 2023-09-09
Variable names appear in table review screen


miabonardi:
<img width="643" alt="Screen Shot 2023-09-09 at 4 06 59 PM" src="https://github.com/SuffolkLITLab/docassemble-ALToolbox/assets/65858862/a5bcf16b-ed86-4006-abe6-6dee57edcbec">

https://github.com/SuffolkLITLab/docassemble-ALToolbox/blob/56e54a9a0fa4ed0cf3c91d63cd4124bd83ef7db5/docassemble/ALToolbox/data/questions/al_income.yml#L450-L459

BryceStevenWilley:
This is intentional; the `table` code is https://github.com/SuffolkLITLab/docassemble-ALToolbox/blob/3ffaedb8d08b0822709801bf2c96505d8648e27d/docassemble/ALToolbox/data/questions/al_income.yml#L149-L159

In the ALItemizedValueDict, we don't know if this is for a job, or an income, etc. You should overwrite the table to use your specific object's terms. Take a look at how it's done in the [Affidavit](https://github.com/SuffolkLITLab/docassemble-ALAffidavitOfIndigency/blob/5c908e47a06ee941c2bbe8395e5292eb3d2dfcf5/docassemble/ALAffidavitOfIndigency/data/questions/affidavit_of_indigency_supplement_body.yml#L885-L897).

We do remove the underscores elsewhere, I can do that easily. But it'll still be lower case and not using the same terms that you are likely using in your interview

PR #234, opened 2023-12-14
Get rid of underscores in ItemizedValueDict table

BryceStevenWilley:
Can't do much more, but we can kind make them look not like variables|

Need to test, I'll merge once I have

#191

### DIFF
--- a/docassemble/ALToolbox/data/questions/al_income.yml
+++ b/docassemble/ALToolbox/data/questions/al_income.yml
@@ -151,7 +151,7 @@ table: x.table
 rows: x
 columns:
   - Type: |
-      str(row_index).replace("_", " ").lower()
+      str(row_index).replace("_", " ").lower().capitalize()
   - Amount: |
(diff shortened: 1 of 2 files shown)


docassemble/ALToolbox/data/questions/al_income.yml
```diff
@@ -151,7 +151,7 @@ table: x.table
 rows: x
 columns:
   - Type: |
-      row_index
+      str(row_index).replace("_", " ").lower()
```
nonprofittechy, line 154:
```suggestion
      str(row_index).replace("_", " ").lower().capitalize()
```

nonprofittechy (changes requested):
Maybe add in an initial capital?

nonprofittechy (approved):

BryceStevenWilley:
Tested and it works fine. Also added a review screen to the demo to easy check on the out of the box situation for things like this

PR #241, opened 2024-01-09
Normalize date creation for calculations and displays

plocket:
Locale wasn't really the problem. The code just needed to be consistent about using local time vs. UTC time. It lines up now.


nonprofittechy (approved):
Looks like an elegant fix. Thank you!

PR #258, opened 2024-06-17
Testing publish action

BryceStevenWilley:


BryceStevenWilley:
Woo, works!

Was testing some changes from https://github.com/SuffolkLITLab/ALActions/pull/26 to make sure that it didn't affect things that didn't use numpy as directly, looks like it does.

PR #264, opened 2024-07-30
Add a few utility functions for DOCX templates - handle None values and Other for checkboxes and radio buttons

### LINKED ISSUE #263: Add a "notnone" function?
```
def notnone(val):
    if val is None:
       return ""
    return val
```

nonprofittechy:
Fix #263

This pull request introduces three new utility functions to enhance template handling and prevent unwanted placeholder text (e.g., "None" or "Other") from appearing in the output. These functions are designed to help write more readable template code.

### none_to_empty

```python
def none_to_empty(val: Any):
    """If the value is None or "None", return a DAEmpty value. Otherwise return the value."""
```
* Purpose: Replaces None or the string "None" with a DAEmpty value to ensure that these placeholders do not appear in the final output.
* Use Case: Particularly useful when handling optional fields, such as radio buttons that may not be selected.

### option_or_other

```python
def option_or_other(
    variable_name: str, other_variable_name: Optional[str] = None
) -> str:
    """If the variable is set to 'Other', return the value of the 'other' variable. Otherwise return the value of the variable."""
```

* Purpose: Handles the scenario where a variable is set to "Other" by returning the corresponding "other" variable's value.
* Use Case: Useful in dropdowns or radio buttons where "Other" is an option, ensuring the actual value provided by the user is displayed instead of the placeholder.

### true_values_with_other

```python
def true_values_with_other(
    variable_name: str, other_variable_name: Optional[str] = None
) -> List[str]:
    """Return a list of values that are True, with the value of the 'other' variable appended to the end of the list."""
```
* Purpose: Collects all True values from a variable, replacing any instances of "Other" with the actual value of the corresponding "other" variable.
* Use Case: Ideal for handling multi-select fields where "Other" might be an option, ensuring the specific user input is included in the results.

### Test case
[test_new_functions.docx](https://github.com/user-attachments/files/16432658/test_new_functions.docx)

```yaml
---
modules:
  - .misc
---
mandatory: True
question: |
attachment:
  docx template file: test_new_functions.docx
---
question: |
  Answer some Qs
fields:
  - Pick one: selected_option
    choices:
      - Option A
      - Option B
      - Other
  - Other: selected_option_other
    show if:
      variable: selected_option
      is: Other
  - Pick a few: checkbox_test
    datatype: checkboxes
    choices:
      - A
      - B
      - C
      - other
  - Other: checkbox_test_other
    show if: checkbox_test["other"]
  - Pick yes no or leave blank: possibly_empty_value
    datatype: radio
    required: False
    choices:
      - Something
      - Other
```

### DIFF
--- a/docassemble/ALToolbox/misc.py
+++ b/docassemble/ALToolbox/misc.py
@@ -1,36 +1,41 @@
-from typing import Dict, List, Optional, TypedDict, Union
+from typing import Any, Dict, List, Optional, TypedDict, Union
 
 from base64 import b64encode
 from decimal import Decimal
 import docassemble.base.functions
 from docassemble.base.util import (
-    defined,
-    value,
-    showifdef,
-    space_to_underscore,
     action_button_html,
     Address,
-    word,
+    DADict,
+    DAEmpty,
+    defined,
+    showifdef,
+    space_to_underscore,
     user_has_privilege,
+    value,
+    word,
 )
 import re
 
 __all__ = [
-    "shortenMe",
-    "thousands",
-    "tel",
-    "fa_icon",
-    "space",
-    "yes_no_unknown",
-    "number_to_letter",
-    "collapse_template",
-    "tabbed_templates_html",
-    "review_widget",
-    "sum_if_defined",
     "add_records",
-    "output_checkbox",
-    "nice_county_name",
     "button_array",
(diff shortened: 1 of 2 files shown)


samglover (approved):
LGTM

PR #268, opened 2024-12-04
Add 'include_a_year' helper function for validation of free-text date fields

nonprofittechy:
Tests are currently not passing because of a breaking change in Docassemble 1.6 that only affects the test, not the runtime (tests aren't able to discover the docassemble.base package).


samglover (approved):
LGTM

Issue #286, opened 2025-09-03
make lists of attributes in documentation for al_income easier to read


ekressmiller:
Apologies if this is not the right place for this.

I was using some al_income objects the other day and noticed that in the [documentation on the Suffolk lit lab page](https://assemblyline.suffolklitlab.org/docs/components/ALToolbox/al_income), the lists of attributes are quite difficult to scan. It would be helpful if each attribute started on a new line or something thing like that.

<img width="995" height="218" alt="Image" src="https://github.com/user-attachments/assets/6df8394a-382b-4651-b191-6b2f28bc0739" />

nonprofittechy:
Fixed in #294

BryceStevenWilley:
Yeah, I missed #294, but looks like the attributes need to be indented, the type wrapped in parenthesis (not curly braces), and have a colon between the name/type and the description.

Something like this works locally:

```
    Attributes:
        .value (str | float | Decimal): A number representing an amount of money accumulated during
            the `times_per_year` of this income.
        .times_per_year (float | Decimal): Represents a number of the annual frequency of
            the income. E.g. 12 for a monthly income.
        .source (str): (Optional) The "source" of the income, like a "job" or a "house".
        .display_name (str): (Optional) If present, will have a translated string to show the
            user, as opposed to a raw english string from the program
```

I'll try to make a PR to fix that part soon.

nonprofittechy:
@BryceStevenWilley no worries, I'm making the PR now :)

I was focusing on the bigger formatting challenge and missed this particular one.

PR #299, opened 2025-09-15
Better use of Google's official docstring format

nonprofittechy:
Correct fix for #286


BryceStevenWilley (approved):
LGTM for the attribute fixes. The examples changes didn't really fix much, but that can be fixed later probably.

Image

PR #305, opened 2025-10-30
Add new extract_fields_from_file function, with examples

nonprofittechy:
1. Adds dependency on Microsoft Markitdown for file conversion, allowing to handle XLSX, DOCX, etc.
2. when the filetype is PDF, uses OpenAI vision API, which is more accurate


nonprofittechy:
Going to merge this; it's a minimal, tested, new feature. But comments welcome.

PR #347, opened 2026-06-15
Switch copy button to use normal text instead of a text input field

### LINKED ISSUE #346: Copy Link input missing label
> On the "Share this website" screen, the text input containing the URL for users to copy does not have an associated label.

> Either add an accessible label to the input (using "aria-label" such as "Share link" or associating it with the visible "You can copy and share this link" text via "aria-labelledby"), or replace the readonly input with a non-form element like a "&lt;span&gt;" since the user does not type into this field. See WebAIM's guidance on form labeling: https://webaim.org/techniques/forms/controls

(See A11Y-09)

nonprofittechy:
Fix #346 

Makes it look pretty much exactly the same as before, except a slight difference on long/scrollable inputs:

image

1. Converts into a `span` element
2. This should now be accessible to  a screen reader

### DIFF
--- a/docassemble/ALToolbox/data/static/copy_button.js
+++ b/docassemble/ALToolbox/data/static/copy_button.js
@@ -2,12 +2,35 @@ $(document).on('daPageLoad', function () {
   try {
     var copy_nodes = document.getElementsByClassName('al_copy');
 
+    function copyText( text ) {
+      if ( navigator.clipboard && navigator.clipboard.writeText ) {
+        return navigator.clipboard.writeText( text );
+      }
+
+      return new Promise(function( resolve, reject ) {
+        try {
+          var temp_input = document.createElement('textarea');
+          temp_input.value = text;
+          temp_input.setAttribute('readonly', '');
+          temp_input.style.position = 'absolute';
+          temp_input.style.left = '-9999px';
+          document.body.appendChild( temp_input );
+          temp_input.select();
+          document.execCommand('copy');
+          document.body.removeChild( temp_input );
+          resolve();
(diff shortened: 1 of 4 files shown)


docassemble/ALToolbox/data/static/copy_button.js
```diff
@@ -25,13 +48,15 @@ $(document).on('daPageLoad', function () {
...
-          $tooltip_inert.hide()
-          $tooltip_active.show()
-          input.blur();
+          copyText( text_to_copy ).then(function() {
+            $tooltip_inert.hide()
+            $tooltip_active.show()
+          }).catch(function( error ) {
+            console.warn( 'Error in AL copy button click' );
+            console.warn( error );
+          });
```
BryceStevenWilley, line 59: don't think we need two nested try catches here?

BryceStevenWilley (approved):
Looks good to me! Confirmed that the accessibility issues caught by Lighthouse before pass now, and confirmed that the screen reader doesn't list the full text of the thing being copied as the form element now, which it was before.

docassemble/ALToolbox/data/static/copy_button.js
plocket: I'm trying to find details about this in js documentation, but this separate `catch` may be necessary here to avoid an uhandled exception-type error. I believe errors in async code, like the Promise here, otherwise go straight to the browser.

Alternatively, we could turn the function that contains this code into an async function and switch the code here to `await copyText(...`

Here's a stack overflow discussion: https://stackoverflow.com/a/68103935/14144258

PR #348, opened 2026-06-15
Change conversion method to convert_local for PDFs

nonprofittechy:
Following the advice of https://github.com/microsoft/markitdown/commit/a51f725d7ff4cdfe3bb6ad2ce2c04d98bf5f1f00, restrict the **default** conversion function to be the local version, and not the more general one.


nonprofittechy:
Just saw this in-progress branch. ~Good idea; but going to make this an overridable default; can think of good reasons to allow remote access, and I don't know if it's more dangerous than allowing an untrusted upload. but it shouldn't be manipulable unintentionally.~

Actually, looking at how the function is built, it shouldn't need to have access to a remote endpoint. The parameter in question must be a DAFile.

PR #351, opened 2026-06-25
[codex] Make times per year lookup robust

nonprofittechy:
## Summary

Makes `times_per_year()` tolerate cadence rows in either `[frequency, label]` or `[label, frequency]` order and return a string label consistently.

## Why

The integrated fee waiver path exposed a row-order mismatch that could return an integer to template code expecting a label.

## Validation

Locally deployed with the integrated MATC 1A divorce package set in the docassemble Docker container and ran the API driver end to end to the final download screen.

### DIFF
--- a/docassemble/ALToolbox/al_income.py
+++ b/docassemble/ALToolbox/al_income.py
@@ -113,7 +113,9 @@ def times_per_year(
     try:
         for row in times_per_year_list:
             if float(times_per_year) == float(row[0]):
-                return row[1].lower()
+                return str(row[1]).lower()
+            if float(times_per_year) == float(row[1]):
+                return str(row[0]).lower()
         return (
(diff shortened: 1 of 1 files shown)


BryceStevenWilley:
I don't think this is a good PR, and that the fix should be applied whereever that times per year list is defined (I can't find where it's used), but I do get that python typing doesn't really extend to YAML yet, and it's not too much code to add.

nonprofittechy:
I had Codex run in a loop to test the 1A divorce packages, and I said it could open draft PRs to solve some of the bugs. I should have restricted its PR opening scope to just the 1A divorce packages, but I didn't. This is one of those stray PRs. Agreed that this isn't the right place to fix mis-use of the existing code.

PR #353, opened 2026-06-29
Fix JsonStorage imports for docassemble 1.10

### LINKED ISSUE #352: Replace removed docassemble.webapp.jsonstore import for docassemble 1.10.0
## Summary

`docassemble/ALToolbox/save_input_data.py` imports from a private docassemble module that no longer exists in docassemble 1.10.0:

```py
from docassemble.webapp.jsonstore import JsonStorage

try:
    from docassemble.webapp.jsonstore import JsonDb
except ImportError:
    from docassemble.webapp.jsonstore import db

    JsonDb = db.session
```

In current `jhpyle/docassemble` 1.10.0, `docassemble.webapp.jsonstore` is gone. The JSON storage model/helper code was moved under `docassemble.webapp.jsonstorage`.

## Why this breaks

This is an import-time failure: any code path that imports `docassemble.ALToolbox.save_input_data` under docassemble 1.10.0 will fail before `save_input_data()` can run.

## Proposed fix

The direct replacement appears to be:

```py
from docassemble.webapp.jsonstorage.helpers import JsonStorage
from docassemble.webapp.extensions import db

JsonDb = db.session
```

The existing write code can then stay mostly the same:

```py
JsonDb.add(new_entry)
JsonDb.commit()
```

## Test suggestion

Add a simple import smoke test against docassemble 1.10.0:

```bash
python - <<'PY'
import docassemble.ALToolbox.save_input_data
PY
```

A functional test should call `save_input_data()` inside a docassemble request/app context and verify that a `JsonStorage` row is created.

## Notes

This looks like the smallest and most mechanical of the 1.10.0 import breakages I found, so it is a good candidate for a quick PR.

nonprofittechy:
## Summary

Fixes the removed `docassemble.webapp.jsonstore` private import for docassemble 1.10 while keeping the old import path as a fallback for older docassemble versions.

The compatibility order is:

1. Try the new docassemble 1.10 locations:
   - `docassemble.webapp.jsonstorage.helpers.JsonStorage`
   - `docassemble.webapp.extensions.db.session`
2. Fall back to the pre-1.10 `docassemble.webapp.jsonstore` path.

Fixes #352.

## Testing

Not run here; this should be smoke-tested in both environments:

```bash
python - <<'PY'
import docassemble.ALToolbox.save_input_data
PY
```

A fuller test should call `save_input_data()` inside a docassemble app/request context and confirm the row is written.

### DIFF
--- a/docassemble/ALToolbox/save_input_data.py
+++ b/docassemble/ALToolbox/save_input_data.py
@@ -12,11 +12,11 @@ except ImportError:
     from docassemble.webapp.jsonstore import JsonStorage
 
     try:
-        from docassemble.webapp.jsonstore import JsonDb
+        from docassemble.webapp.jsonstore import JsonDb  # type: ignore[no-redef]
     except ImportError:
         from docassemble.webapp.jsonstore import db
 
-        JsonDb = db.session
+        JsonDb = db.session  # type: ignore[no-redef]
 
 
 __all__ = ["save_input_data"]
--- a/docassemble/ALToolbox/save_input_data.py
+++ b/docassemble/ALToolbox/save_input_data.py
@@ -2,14 +2,10 @@ from docassemble.base.generate_key import random_alphanumeric
 from docassemble.base.functions import get_current_info
 from docassemble.base.util import DADict
 from typing import Dict, List, Any, Optional
-from docassemble.webapp.jsonstore import JsonStorage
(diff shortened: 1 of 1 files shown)


nonprofittechy:
Tested, works on both old/new DA versions

PR #366, opened 2026-09-19
Let chat_completion send images

### LINKED ISSUE #364: chat_completion cannot send images: three blockers, and a proposed opt-in parameter
## What I was doing

Adding an AI alt-text pass to the PDF accessibility workshop in ALDashboard: send a page image to a vision model and get back one sentence of alternative text. `chat_completion` is the natural place to do that, and the OpenAI-compatible endpoints it talks to already accept multimodal content — a `content` that is a **list of parts** rather than a string:

```python
{"role": "user", "content": [
    {"type": "text", "text": "What is its alternative text?"},
    {"type": "image_url", "image_url": {"url": "data:image/png;base64,..."}},
]}
```

That works today if you pass it through `messages=`, but three things in `llms.py` get in the way.

## The three blockers

**1. `messages` is typed as string-only content.** `chat_completion(..., messages: Optional[List[Dict[str, str]]] = None)` (llms.py:459). A multimodal message is `Dict[str, Any]`, so every caller has to `cast` or `# type: ignore` at the call site. Runtime is fine; the annotation just doesn't describe what the endpoint accepts.

**2. `json_mode` assumes content is a string and raises on a list.** llms.py:501:

```python
if (messages and json_mode
        and not any("json" in message["content"].lower() for message in messages)):
```

`list.lower()` → `AttributeError`. So JSON mode and images are mutually exclusive, and the failure is an unhelpful traceback rather than a clear error. (For alt text I don't want JSON, so I could work around this — but "describe these five images and return a JSON array" is a reasonable thing to want.)

**3. Token counting stringifies the payload.** llms.py:554:

```python
token_count = len(encoding.encode(str(messages)))
```

With an image inlined as a base64 data URI, this tokenises the base64 itself. A modest 768px PNG runs a few hundred KB, which becomes a few hundred thousand characters and can trip `max_input_tokens` (default 128000) even though the *provider* bills an image at a fixed, much smaller cost. It also spends real time running tiktoken over base64.

This is the one that actually breaks things rather than merely annoying the type checker.

## Proposed shape

Rather than loosening `messages` and asking every caller to assemble parts by hand, add an opt-in parameter, so everything else stays exactly as it is:

```python
def chat_completion(
    ...,
    images: Optional[List[Union[str, bytes]]] = None,
    image_detail: Optional[Literal["low", "high", "auto"]] = None,
) -> ...
```

- Each entry is raw image bytes, a `data:` URI, or an `http(s)` URL. Bytes get base64-encoded and the media type sniffed.
- When `images` is set, the **last user message** becomes multimodal: its existing text becomes a `{"type": "text"}` part, followed by one `{"type": "image_url"}` part per image. Callers that pass `user_message=` rather than `messages=` keep working unchanged.
- The `json_mode` check reads text parts only, so JSON mode and images can be combined.
- Token counting counts the text parts and adds a flat nominal cost per image rather than tokenising base64.

Nothing changes for any existing caller: with `images=None` the assembled messages are byte-for-byte what they are today.

## Why a parameter rather than just widening the type

Three reasons, all about keeping the failure modes visible:

- The token-count and `json_mode` fixes have to happen *somewhere*. If callers hand-assemble parts, every caller has to know about both.
- Sending pixels to a third party is a decision worth naming in the signature. `images=` at the call site is greppable; a shape buried in a `messages` list is not.
- It leaves room to do the right thing per provider later (Anthropic's image blocks differ from OpenAI's) without changing any caller.

## Happy to send the PR

I have this working against a vision endpoint from ALDashboard and will open a PR implementing the above, with tests covering: images appended to an existing `messages` list, images with `user_message=`, `json_mode` combined with images, and token counting not tokenising base64. Say the word if you'd prefer a different shape and I'll rework it.

Context: SuffolkLITLab/docassemble-ALDashboard `pdf-accessibility-workshop`, `describe_images_with_ai` in `pdf_accessibility.py`.

nonprofittechy:
Closes #364.

Adds an opt-in `images` parameter rather than widening `messages`, so nothing changes for any existing caller — with `images=None` the assembled messages are byte for byte what they were, and there is a test asserting exactly that.

```python
chat_completion(
    system_message="Describe the image in one sentence.",
    user_message="What is this?",
    images=[png_bytes],
    image_detail="low",
)
```

Each entry is raw image bytes, a `data:` URI, or an `http(s)` URL. Bytes are base64-encoded with the media type sniffed from the magic number. Only the **last user message** changes shape, into the list of parts the endpoint expects; every other message is untouched, and the caller's own list is not mutated.

## What had to move

Three things in `llms.py` got in the way of a picture, all described in #364:

| | before | now |
|---|---|---|
| `messages` typing | `List[Dict[str, str]]`, so a multimodal message needed a `cast` at every call site | callers keep writing ordinary string-content messages; the shape is assembled here |
| `json_mode` | `message["content"].lower()` → `AttributeError` on list content | reads the text parts, so JSON mode and images can be combined |
| token counting | `encoding.encode(str(messages))` tokenised the base64 itself | text parts counted as before, plus a flat estimate per image |

The token one was the only functional blocker: a 768px photo runs a few hundred KB, which became a few hundred thousand characters and could trip `max_input_tokens` on an input the endpoint would have accepted. Moderation had the same problem — it was sending base64 to the moderation endpoint — and now sees the words too.

The per-image estimates are OpenAI's published figures (85 tokens at `detail="low"`, 1105 otherwise). They are a guard rail, not billing, and are named as estimates in the code.

## Tests

`test_llms_images.py`, 14 tests: media-type sniffing for PNG/JPEG/GIF/WebP, refusal of anything that is not bytes or a URL, attaching to an existing `messages` list, the case with no user message to attach to, `detail` pass-through, several images on one message, `json_mode` alongside images, a 2 MB image not tripping the input limit, and no-images-means-no-change.

They import `llms` inside the test bodies rather than at module level: `llms` reads the docassemble configuration at import, and the `conftest.py` fixture that stands in for a server is session-scoped, so it is not registered yet during collection. Worth knowing for anyone else writing the first test against this module.

`black --check` clean, and the existing suite still passes (94 tests).

## One thing noticed and left alone

`openai_base_url` is given a default at llms.py:587, before the client check at llms.py:618:

```python
if openai_base_url:
    openai_client = None  # Always override client in this circumstance
```

Since the url is now always truthy, that branch always fires and an `openai_client=` argument is **always** discarded — the parameter only works if the module-level `client` happens to exist. It looks like the guard predates the base-url default. Out of scope here, so the tests patch the module-level client instead; happy to open a separate issue if you would like it tracked.

## Where this is used

SuffolkLITLab/docassemble-ALDashboard `pdf-accessibility-workshop`: `describe_images_with_ai` sends a downscaled page image to a vision model and asks for one sentence of alternative text, for images that are still pictures after OCR has handled the scans. It currently hand-assembles the parts and casts; with this merged it can just pass `images=`.


nonprofittechy:
The failing `build` check is not from this PR.

`da_build` is pinned to `SuffolkLITLab/ALActions/da_build@main`, and that action has picked up DOCX accessibility checks since the last run here a month ago. Every failure in the log is in `.docx` templates — `WA557` heading structure starts at level 2, `IA561` missing title metadata, `IA565` empty paragraphs used for spacing — none of which this branch touches. The diff is two Python files.

To be sure rather than assume, I re-ran the last `main` build (run `32322762644`, commit `4e4ecc0`, unmodified) against the current action: it **also fails**, with the same class of DOCX findings plus an `EG105`. So this is a pre-existing condition on `main` that only became visible when something ran again, not a regression here.

Every other check passes, including **Run python only unit tests** and **formatting**.

Worth fixing on `main` separately — and there is a nice irony in an accessibility toolkit's own templates failing an accessibility lint.

